PR #14, opened 2022-06-04
solution

MarkoKhodan:


app/classes/Customer.py
```diff
@@ -0,0 +1,69 @@
+import json
+from dataclasses import dataclass
+from math import sqrt
+from datetime import datetime
+
+with open("../app/config.json", "r") as f:
```
IvanRamyk, line 6: Wrong path to the config, the tests are failes

app/main.py
```diff
@@ -1,3 +1,40 @@
+import json
+from app.classes.Customer import Customer
+from app.classes.Shop import Shop
+from app.classes.Car import Car
+
+
 def shop_trip():
-    # write your code here
-    pass
+    with open("../app/config.json", "r") as f:
```
IvanRamyk, line 8: Same here

IvanRamyk (changes requested):

app/classes/Customer.py
MarkoKhodan: this is the only way, how I can run tests in venv, I tried "app.config.json" and "config.json" and both cases doesn't work

IvanRamyk (approved):
Good job!

PR #15, opened 2022-06-07
Solution

rotsen18:


app/customers.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Customer:
+    def __init__(self, user_info, fuel_price):
+        self.name = user_info["name"]
+        self.product_cart = user_info["product_cart"]
+        self.location = user_info["location"]
+        self.money = user_info["money"]
+        self.car = user_info["car"]
+        self.fuel_price = fuel_price
```
Magus6784, line 11: I don't think fuel price is correct attribute for Customer. You know, fuel price not connect to Customer. If fuel price changed you need to change all your Customer

app/customers.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def fuel_costs(self, shop):
+        distance = math.dist(shop.location, self.location)
+        return self.fuel_price * distance * self.car["fuel_consumption"] / 100
+
+    def pay_for_trip(self, amount):
+        self.money -= amount
+        return self.money
```
Magus6784, line 19: You don't need to return self.money. You can just use customer.money in main/$20

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    for shop in shops:
+        costs = customer.trip_costs(shop)
+        print(f"{customer.name}'s trip to the {shop.name} costs {costs}")
+        shops_cost[costs] = shop
+    shops_cost = dict(sorted(shops_cost.items()))
+
+    for costs, shop in shops_cost.items():
+        if costs <= customer.money:
+            return shop, costs
+    return None, None
```
Magus6784, line 39: maybe its better to create method for Customer.
If you create method for trip in Customer, why choose shop not in class too?

Magus6784 (changes requested):

Magus6784:
Also i think better name for files is just. "customer" and "shop". Like in tasl elfes and dwarfes we use only "Elf" "Dwarf"

Magus6784:
And a little more
"When the customer arrives at the shop his location should equal to shop location."

TPolina (approved):
well done!

PR #16, opened 2022-06-08
Solution

YuriiKindrat:


app/classes.py
```diff
@@ -0,0 +1,91 @@
...
+                best_shop = shop
+
+        if self.money > min_cost:
+            print(f"{self.name} rides to {best_shop.name}\n")
+            best_shop.purchase(self)
+            self.money -= min_cost
+            self.coming_home()
+        else:
+            print(f"{self.name} doesn't have enough "
+                  f"money to make purchase in any shop")
```
Magus6784, line 61: "When the customer arrives at the shop his location should equal to shop location."

Magus6784 (changes requested):

Magus6784 (changes requested):
Your tamadachi lives in shop now :(

TPolina (approved):
better write each class in separate modules and then combine them in the main module and implement the main logic
creating modules classes, functions is not really good

PR #17, opened 2022-06-09
Solution

Magus6784:


app/customer.py
```diff
@@ -0,0 +1,52 @@
...
+
+@dataclass
+class Customer:
+    name: str
+    product_cart: dict
+    location: list
+    money: float
+    car: dict
+
+    def find_cheapest_trip_to_shop(self, shops: list, fuel_price: float):
```
KatiukhaO, line 14: its good function, good way, maybe, i append the same func and in my code

KatiukhaO (approved):
nice

TPolina (approved):

PR #19, opened 2022-06-09
Solution

KatiukhaO:


.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 ignore = E203, E266
-max-line-length = 79
+max-line-length = 130
```
Magus6784, line 3: If we change flake8, what will be left of our society?

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+
+def create_instance_customers(story: dict):
+    customers = [
+        Customer(
+            name=customer["name"],
+            products_cart=customer["product_cart"],
+            location=customer["location"],
+            money=customer["money"],
+            car=Car(brand=customer["car"]["brand"],
+                    fuel_consumption=customer["car"]["fuel_consumption"])
```
Magus6784, line 18: extra indentation?

app/customer.py
```diff
@@ -0,0 +1,45 @@
+import math
+
+from app.car import GasStation, Car
+from app.shop import Shop
+
+
+class Customer:
+    fuel_price = GasStation(2.4)
```
Magus6784, line 8: why you save fuel_price in Customer
Firstly we take fuel_price from json file
Secondly how fuel_price connect with customer?

Magus6784: Also i don't think we need class for this, but maybe ok

Magus6784 (changes requested):
We can use module dataclasse to init class, but up to you

PR #20, opened 2022-06-09
Solution

darten360:


app/distance.py
```diff
@@ -0,0 +1,11 @@
+class Distance:
+    def __init__(self, start: list, end: list):
+        self._start = start
+        self._end = end
+
+    def distance(self):
+        first_addition = (self._end[0] - self._start[0]) ** 2
```
KatiukhaO, line 7: it s  correct, and i started do this the same, but in method dist is in math module, you could use this

KatiukhaO (approved):
interesting logic of project

app/which_shop.py
```diff
@@ -0,0 +1,96 @@
...
+        ).distance()
+
+        final_price = Shop.cart_price_list(self)
+        price_for_1_km = Shop.price_for_1_km(self)
+        result = final_price + distance * price_for_1_km * 2
+
+        return round(result, 2)
+
+
+class WhichShop(Shop):
```
TPolina, line 45: how does this class relate to objects in real life?
Class Shop is okay. It represents all stores in real life. "АТБ", "СІЛЬПО", and "АШАН" will be instances of this class. But what WhichShop stand for? What will be an instance of it?

TPolina: I recommend to create class shop and class customer. Inside these classes implement functionality relates only to shop or only to customer. Then use these classes in main module with main logic.

app/which_shop.py
```diff
@@ -0,0 +1,96 @@
...
+
+class Shop:
+
+    def __init__(
+            self,
+            customer: dict,
+            shop: dict,
+            dct: dict
+    ):
+        self._customer = customer
```
TPolina, line 13: customer shouldn't be an attribute of Shop instance. Customers come and go. They are not tied to shop

MasterpieceElbow (approved):

PR #21, opened 2022-06-09
Solution

4ndyua:


rotsen18 (approved):
1. Try to keep size each function at least under 10 lines. Maybe split last into something that parse config (and then you won't have code in root of module) and the main logic.
2. There isn't realization of changing customer location but I'm not sure how it is imoirtant in this task.

TPolina (approved):

PR #22, opened 2022-06-09
'shop_trip'

CheshkoSergey:


NOSSUDAK (approved):
I would suggest you to use OOP for tasks like this :)

PR #23, opened 2022-06-10
Develop

Kiawami:


app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    customer.change_location(shop.location_x, shop.location_y)
+    customer.pay(trip_price)
+
+    shop.give_check(customer.name, customer.product_cart)
+    purchase_price = customer.count_purchase_price(shop)
+    customer.pay(purchase_price)
+
+    customer.pay(trip_price)
+    customer.go_home()
+    customer.change_location(home_location[0], home_location[1])
```
DmytroSvirsa, line 48: It looks like repeating yourself, maybe better to put it in `go_home` and create something like `go_to_shop`, and change location inside&

DmytroSvirsa (approved):

PR #24, opened 2022-06-10
Solution

Serhii-Leonenko:


app/main.py
```diff
@@ -1,3 +1,125 @@
...
+        trip_cost_data_dict["products"][product] = (amount, product_cost)
+        total_product_cost += product_cost
+
+    trip_cost_data_dict["total_product_cost"] = total_product_cost
+    trip_cost_data_dict["the_whole_cost"] = total_product_cost + fuel_cost
+
+    return trip_cost_data_dict
+
+
 def shop_trip():
```
KatiukhaO, line 59: i think , this function is very long, may split her? for exemple, from $99-112 line, this can be in class Shop, like method "print a check", what do you thinking?

KatiukhaO (changes requested):

app/main.py
Serhii-Leonenko: I have decided to implement all prints in one fuction, classes used only for creating instances . If you think that is bad implementaition, i will remake it

KatiukhaO: yes, i think, should, refactor this,

Serhii-Leonenko: ok

KatiukhaO: i think, print a check its responsibility class Shop

KatiukhaO (approved):

app/main.py
Serhii-Leonenko: agree

Serhii-Leonenko: i will refector

TPolina (approved):

PR #25, opened 2022-06-10
Solution

SergeyDev1996:


app/main.py
```diff
@@ -1,3 +1,80 @@
...
+        cost = get_cost_of_ride_to_the_store(customer, shop)
+        cost_of_purchases.append(cost)
+        stores[shop_name] = cost
+        print(f"{name}'s trip to the {shop_name} costs {cost}")
+    max_cost = max(cost_of_purchases)
+    cheapest_store = min(stores, key=stores.get)
+    for shop in shops:
+        if shop["name"] == cheapest_store:
+            cheapest_store = shop
+    if money >= max_cost:
```
rotsen18, line 48: I'm not sure but if we don't have the money for the most expensive trip then we don't go at all? Will your program work when we have 10$ and trips costs 5, 8, 15 for example?

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+        print(f"{name} rides to {cheapest_store['name']}\n")
+        purchase_in_the_store(customer, cheapest_store)
+    else:
+        print(f"{name} doesn't have enough money to make purchase in any shop")
+
+
+def purchase_in_the_store(customer: dict, shop: dict):
```
rotsen18, line 55: There no implementation of changing customer location but I don't now how it is important for this task.

rotsen18 (changes requested):

app/main.py
SergeyDev1996: Checked it. We only have this: "When the customer arrives at the shop his location should equal to shop location." As far as I understand, this is just an instruction on how to calculate the distance, and doesn't looks it should be implemented in the code.

app/main.py
SergeyDev1996: You are right

rotsen18 (approved):

TPolina (approved):

PR #29, opened 2022-06-10
solution

DmytroSvirsa:


app/customers.py
```diff
@@ -0,0 +1,73 @@
...
+        self.money = money
+        self.car = car
+        self.home = self.location
+
+    def get_trip_cost(self, shop):
+        distance = sum(
+            [(x - y) ** 2 for x, y in zip(
+                shop.location,
+                self.location
+            )]) ** 0.5
```
Magus6784, line 23: better use math.dist

app/customers.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class Customer:
+    def __init__(self, name, product_cart, location, money, car):
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.money = money
+        self.car = car
+        self.home = self.location
```
Magus6784, line 16: Why not just location?

Magus6784: Also you can try rewrite with @dataclass

Magus6784 (changes requested):

Magus6784 (approved):

TPolina (approved):

PR #30, opened 2022-06-10
'Solution'

Aleksei-Isaev:


app/main.py
```diff
@@ -1,3 +1,77 @@
...
+    with open("app/config.json", "r") as data_json:
+        user_data = json.load(data_json)
+
+    fuel_price_data = user_data["FUEL_PRICE"]
+    customers_data = user_data["customers"]
+    shops_data = user_data["shops"]
+
+    for customer in customers_data:
+        print(f"{customer['name']} has {customer['money']} dollars")
+        overall_prices = price_for_trip(customer, shops_data, fuel_price_data)
```
rotsen18, line 16: As for me, when you count and return several values, it would be good to call in the plural (for example prices_for_trips). I was a little confused why the function is price_for_trip (singular) when we have overall_prices (plural) variable.

rotsen18 (approved):

PR #31, opened 2022-06-10
'Solution'

YuraBeseda:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        else:
+            print(f"{customer.name}"
+                  f" doesn't have enough money to make purchase in any shop")
+
+
+def create_customers_and_shops():
+    with open(BASE_DIR / "config.json", "r") as f:
+        data = json.load(f)
+
+    FUEL_PRICE = data["FUEL_PRICE"]
```
rotsen18, line 32: Upper case is good for constants, but there is just local variable. The same in return.

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            print(f"{customer.name} now has "
+                  f"{customer.money - min(customer.trips.values())} dollars\n")
+        else:
+            print(f"{customer.name}"
+                  f" doesn't have enough money to make purchase in any shop")
+
+
+def create_customers_and_shops():
+    with open(BASE_DIR / "config.json", "r") as f:
+        data = json.load(f)
```
rotsen18, line 30: Try to use more descriptive names for variables.

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        print(f"{customer.name} has {customer.money} dollars")
+        customer.find_better_price(shops, fuel_price)
+        for shop, price in customer.trips.items():
+            print(f"{customer.name}'s trip to the {shop} costs {price}")
+        if customer.money > min(customer.trips.values()):
+            print(f"{customer.name} rides to {customer.cheapest_shop.name}\n")
+            customer.do_shopping()
+            print(f"{customer.name} rides home")
+            print(f"{customer.name} now has "
+                  f"{customer.money - min(customer.trips.values())} dollars\n")
```
rotsen18, line 22: You are finding min costs trip so many times in app. If you want you may try to do some refactoring

rotsen18 (changes requested):

PR #36, opened 2022-06-12
'Solution'

Korenely:


rotsen18 (commented):
1. Try to keep size each function at least under 10 lines. Maybe split last into something that parse config (and then you won't have code in root of module) and the main logic.
2. There isn't realization of changing customer location but I'm not sure how it is imoirtant in this task.

app/main.py
```diff
@@ -1,3 +1,61 @@
+import json
+import math
+import datetime
+
+with open("app/config.json") as f:
+    data = json.load(f)
```
rotsen18, line 6: Try to use more descriptive names for variables

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    fuel_price = data['FUEL_PRICE']
```
rotsen18, line 7: Use double quotes "" . Check other code for this mistake.

rotsen18 (changes requested):

rotsen18 (approved):

TPolina (approved):

PR #48, opened 2022-07-07
Solution

pavlejviki:


app/shop.py
```diff
@@ -0,0 +1,50 @@
...
+        return round(fuel_expenses, 2)
+
+    def cost_of_trip(self, customer, fuel_price):
+        travel_cost = self.distance_to_shop(customer, fuel_price)
+        shopping_cost = self.shopping(customer)
+        self.total_cost = travel_cost + shopping_cost
+        return self.total_cost
+
+    def __lt__(self, other):
+        return self.total_cost < other.total_cost
```
Abnormaltype, line 50: why do you overload this?

Abnormaltype (changes requested):

app/shop.py
pavlejviki: Apparently it wasn't a good idea. Fixed it

Abnormaltype (approved):
well done!

PR #57, opened 2022-08-14
Solution

andrii-padalko:


app/main.py
```diff
@@ -1,3 +1,117 @@
+import json
+
```
DmytroSvirsa, line 2: the blank line does not need, all imports from the same category -built in libraries

app/main.py
```diff
@@ -1,3 +1,117 @@
...
+import datetime
+
```
DmytroSvirsa, line 4: the blank line does not need, all imports from the same category -built in libraries

app/main.py
```diff
@@ -1,3 +1,117 @@
...
-    # write your code here
-    pass
+    # with open("config.json", "r") as f:
+    with open("app/config.json", "r") as f:
+        data = json.load(f)
+
+    fuel_price = data["FUEL_PRICE"]
+
+    cust_lst = [data["customers"][i]["name"]
+                for i in range(len(data["customers"]))]
```
DmytroSvirsa, line 16: too complicated, why don't `[customer["name"] for customer in data["customers"]]`

DmytroSvirsa (changes requested):
Rewrite using `classes`, split functionality to diferrent modules, try no not spam `range(len(data["customers"]))`

DmytroSvirsa (changes requested):
Great, but let's take the last step, good variable names are very important, and even more so in large projects, more important than the length of the line, I'm sure you can use the rules about line separation, and achieve beautiful, readable code, with meaningful variable names

DmytroSvirsa (approved):
Well done!

PR #65, opened 2022-09-01
'Solution'

dmitriyKostiuk:


app/main.py
```diff
@@ -1,3 +1,135 @@
...
+        sum_bread = self.bread * other.bread
+        sum_butter = self.butter * other.butter
+        return sum_milk + sum_bread + sum_butter
+
+    def getattr(self, item):
+        return self.__getattribute__(item)
+
+
+class Customer:
+    name: str
```
IvanRamyk, line 39: why it's a class attribute?

app/main.py
```diff
@@ -1,3 +1,135 @@
...
+
+        return self
+
+
+class Product:
+    milk: int = 0
+    bread: int = 0
+    butter: int = 0
+
+    def from_dict(self, data):
```
IvanRamyk, line 22: Usually, such method shouldn't set attributes but call a constructor creating a new instance of the class with the corresponding attributes

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #71, opened 2022-09-09
Task Complete

Glasis9:
What could be the problem?
Outputs everything the same as in the tests

### DIFF
--- a/app/car.py
+++ b/app/car.py
@@ -42,7 +42,7 @@ def trip_price(customer,
     """
     Calculation of purchase costs in different shops with a driveway
     """
-    temp_dict = {}
+    dict_of_shopping_and_fare_prices = {}
     for i in range(len(shops)):
         cost = product_calculation(count_milk,
                                    count_bread,
@@ -51,7 +51,8 @@ def trip_price(customer,
 
         cost_plus_fuel = fare[f'fare_to_the_{shops[i]["name"]}'] + cost
 
-        temp_dict[f'{shops[i]["name"]}'] = round(cost_plus_fuel, 2)
+        dict_of_shopping_and_fare_prices[f'{shops[i]["name"]}'] \
+            = round(cost_plus_fuel, 2)
 
         print(f'{customer["name"]}\'s trip to the {shops[i]["name"]} '
               f'costs {round(cost_plus_fuel, 2)}')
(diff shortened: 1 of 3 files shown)


app/open_json.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+
+with open("config.json") as file:
```
Abnormaltype, line 4:
```suggestion
with open("app/config.json") as file:
```

Abnormaltype (changes requested):

app/open_json.py
Glasis9: Why do you need to specify the full path with pytest, but with normal compilation without app/?

Abnormaltype: it's for the tests

app/car.py
```diff
@@ -0,0 +1,85 @@
...
+              f'costs {round(cost_plus_fuel, 2)}')
+
+    """
+    Choosing a store at a lower price and the amount of money a person has
+    """
+    temp_list = []
+    for value in temp_dict.values():
+        temp_list.append(value)
+
+    for key, value in temp_dict.items():
```
Abnormaltype, line 66: uninformative names for a variables

app/open_json.py
```diff
@@ -0,0 +1,26 @@
+import json
+
+
+with open("app/config.json") as file:
+    config_json = json.load(file)
+
+# with open("config.json") as file:
+#     config_json = json.load(file)
```
Abnormaltype, line 8: remove unnecessary comments

app/shop.py
```diff
@@ -0,0 +1,35 @@
...
+             shop):
+    """
+    Trip to the store (change of coordinates)
+    """
+    trip_to_the_store(customer, shop)
+
+    """
+    Receipt printing
+    """
+    date = datetime(2021, 1, 4, 12, 33, 41).strftime("%d/%m/%Y %H:%M:%S")
```
Abnormaltype, line 20: use datetime.datetime.now()

Abnormaltype (changes requested):

app/shop.py
Glasis9: I did this at first, but then I changed it, I thought that the tests might not pass

Glasis9: I'll fix everything now

Abnormaltype (approved):
Well done!

PR #74, opened 2022-09-16
My Solution

YanPurdenko:


app/receipt.py
```diff
@@ -0,0 +1,66 @@
...
+          "dollars"
+          )
+    print(f"{bread_amount} breads for "
+          f"{bread_amount * Decimal(str(min_shop_offer.products['bread']))} "
+          "dollars"
+          )
+    print(f"{butter_amount} butters for "
+          f"{butter_amount * Decimal(str(min_shop_offer.products['butter']))} "
+          "dollars"
+          )
```
Abnormaltype, line 62: DRY, create a loop for this action

Abnormaltype (changes requested):

app/customer.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass()
```
IvanRamyk, line 4: This way to use `dataclass` decorator is OK but the following way is more common:
```suggestion
@dataclass
```

IvanRamyk (approved):
Well done!

PR #80, opened 2022-09-23
Shop trip solution

Foggy-Cat:


app/main.py
```diff
@@ -1,3 +1,71 @@
...
+from app.home import Home
+from app.customer import Customer
+from app.shop import Shop
+from app.init import init_all_class
+
+
 def shop_trip():
-    # write your code here
-    pass
+    with open("../app/config.json", "r") as info:
```
IvanRamyk, line 11: Change file path in order to pass tests
To make sure that the path is correct run tests using the following command in the terminal:
```
pytest . 
```

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #86, opened 2022-09-28
Solution

wQuelS:


AriohBelskij (approved):
good

app/shop.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Shop:
+    def __init__(self, shop: dict):
+        self.name = shop["name"]
+        self.location = shop["location"]
+        self.products = shop["products"]
+
+    def calculate_cost(self, product_cart: dict) -> int | None:
+        total_cost = 0
+        for item in product_cart:
```
Nattalli, line 12: would it better to  name `product` instead of `item`?

app/shop.py
```diff
@@ -0,0 +1,26 @@
...
+            if item not in self.products:
+                return False
+            total_cost += product_cart[item] * self.products[item]
+        return total_cost
+
+    def print_purchase_receipt(self, name: str, product_cart: dict) -> None:
+        print(datetime.datetime.now().strftime("Date: %d/%m/20%y %H:%M:%S"))
+        print(f"Thanks, {name}, for you purchase!")
+        print("You have bought: ")
+        for item in product_cart:
```
Nattalli, line 22: and here. WDYT?

Nattalli (changes requested):

app/shop.py
wQuelS: Actually, i don't agree. Cuz for me "item" is pretty descriptive name for variable to loop over product_cart. But anyway, i'll fix it:)

Nattalli: but if you do not agree, then you can not fix it. That's why I'm asking for your opinion

Nattalli: Anyway you made a great job!

Nattalli (approved):
Amazing 😻

app/shop.py
wQuelS: > but if you do not agree, then you can not fix it. That's why I'm asking for your opinion

Y're the boss here :))

PR #87, opened 2022-09-29
Solution

AriohBelskij:


wQuelS (approved):

app/trip.py
```diff
@@ -0,0 +1,87 @@
...
+        return dist(self.customer.location, self.shop.location)
+
+    def cost_ride(self) -> float:
+        calculation_cost = self.customer.car / 100 * self.get_distance()
+        return calculation_cost * self.fuel_price
+
+    def cost_product(self) -> int:
+        total_cost = 0
+        for product in self.customer.products:
+            list_of_products = self.customer.products[product]
```
Nattalli, line 23: maybe something like `product_amount` or `product_quantity` would be better? WDYT?

Nattalli (changes requested):

app/trip.py
AriohBelskij: yea i think you are right, i fix it.

Nattalli (approved):
Good job! 😸

PR #93, opened 2022-10-04
Solution

AVyha:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+from app.car import Car
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        data = json.load(file)
+    Car.fuel_price = data["FUEL_PRICE"]
+    customers_list = [0] * len(data["customers"])
```
mltdll, line 12: I can't really see what's the point of doing this instead of creating an empty list and appending customer objects to it. Can you explain?

app/shop.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def products_price(self, product_list: dict) -> float:
+        total = 0
+        for product, amount in product_list.items():
+            total += self.products[product] * amount
+        return total
+
+    def buy_products(self, customer) -> None:
+        # print(f"Date: {datetime.now().strftime('%d/%m/%Y %H:%M:%S')}")
+        # for test we need const time
```
mltdll, line 24: We actually don't: there's mocked datetime in test, but you have to import it in a specific way. Import the whole package and use this: 
```python
datetime.datetime.now().strftime("Date: %d/%m/20%y %H:%M:%S\n")

mltdll (changes requested):

app/main.py
AVyha: brain lag)

mltdll (approved):

Y-Havryliv (approved):
Well done!

PR #94, opened 2022-10-05
Solution for the task

olia-trofymchuk:


mltdll (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file_open:
+        data = json.load(file_open)
+
+    customers = [Customer(customer) for customer in data["customers"]]
+    shops = [Shop(shop) for shop in data["shops"]]
+
+    for i in range(len(customers)):
+        customer = customers[i]
```
Nattalli, line 16: is it the same as `for customer in customers`?

Nattalli (changes requested):
Well done, left 1 comment

app/main.py
olia-trofymchuk: Yes:) I have changed it) thank you)

Nattalli (approved):
Good job!!!

PR #95, opened 2022-10-06
solution

KirillDrago:


app/car.py
```diff
@@ -0,0 +1,6 @@
+class Car:
+    def __init__(self, car: dict) -> None:
+        self.fuel_consumption = car["fuel_consumption"]
```
mltdll, line 3: Yeah, why even bother with those "brand" things? Not like they change anything. :)
It's just a comment. The only reason for reject is the other one.

app/main.py
```diff
@@ -1,3 +1,23 @@
...
+
+        for customer in customer_data:
+            print(f"{customer.name} has {customer.money} dollars")
+            customer.choose_shop(shop_data, fuel_price)
+            if customer.choosen_shop:
+                customer.choosen_shop.print_receipt(customer)
+                print(f"{customer.name} rides home")
+                print(f"{customer.name} now "
+                      f"has {customer.money - customer.best_price} dollars\n")
+            else:
```
mltdll, line 22: This `else` block does exactly nothing. The function terminates anyway.

mltdll (changes requested):

app/main.py
```diff
@@ -1,3 +1,22 @@
...
+
+        for customer in customer_data:
+            print(f"{customer.name} has {customer.money} dollars")
+            customer.choose_shop(shop_data, fuel_price)
+            if customer.choosen_shop:
+                customer.choosen_shop.print_receipt(customer)
+                print(f"{customer.name} rides home")
+                print(f"{customer.name} now "
+                      f"has {customer.money - customer.best_price} dollars\n")
+        return
```
mltdll, line 22: Now, what does this return do? By removing `else` block I meant that you should remove `return` as well. Thr function terminates and returns `None` anyway.

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #105, opened 2022-10-09
Solution

taras-andruschenko:


AriohBelskij (approved):

app/customer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def count_remaining_money(self, all_expenses: float) -> None:
+        amount = round(self.money - all_expenses, 2)
+        print(f"{self.name} now has {amount} dollars\n")
+
+
+def make_list_customers(data: dict) -> list:
+    customers = []
+
+    for i, customer in enumerate(data["customers"]):
```
MasterpieceElbow, line 35: Why do use `enumerate` if you don't use it?

app/customer.py
```diff
@@ -0,0 +1,45 @@
...
+        customers.append(
+            Customer(
+                customer["name"],
+                customer["product_cart"],
+                customer["location"],
+                customer["money"],
+                customer["car"]
+            )
+        )
+    return customers
```
MasterpieceElbow, line 45: You can do this function with a single return statement and list comprehension.

app/shop.py
```diff
@@ -0,0 +1,68 @@
...
+
+    for i, shop in enumerate(data["shops"]):
+        shops.append(
+            Shop(
+                shop["name"],
+                shop["location"],
+                shop["products"]
+            )
+        )
+    return shops
```
MasterpieceElbow, line 68: The same as with `make_customer_list`

MasterpieceElbow (changes requested):

app/shop.py
```diff
@@ -0,0 +1,76 @@
...
+    #
+    # for shop in data["shops"]:
+    #     shops.append(
+    #         Shop(
+    #             shop["name"],
+    #             shop["location"],
+    #             shop["products"]
+    #         )
+    #     )
+    # return shops
```
TPolina, line 68: delete comments

app/main.py
```diff
@@ -1,3 +1,39 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+import app.customer as customer
+import app.shop as shop
+
+
+def shop_trip() -> None:
```
TPolina, line 6: In my opinion, this function is too complicated and needs refactoring. But it's up to you

TPolina (changes requested):

TPolina (approved):

PR #107, opened 2022-10-11
Solution)

Gr1zzZzzly:


app/customers_shops/customers.py
```diff
@@ -0,0 +1,65 @@
...
+            print(f"Thanks, {self.name}, for you purchase!")
+            print("You have bought: ")
+            print(self.shops_price[best_shop]["purchase"][:-1])
+            print(f"Total cost is "
+                  f"{self.shops_price[best_shop]['price']} dollars")
+            print("See you again!\n")
+            self.ride_home(best_shop)
+        else:
+            print(f"{self.name} doesn't have enough "
+                  f"money to make purchase in any shop")
```
misha-doroshenko, line 59: You can replace this if/else statement with:
if self.money < products_price:
    print("Not enough money")
    return

app/main.py
```diff
@@ -1,3 +1,39 @@
...
-    # write your code here
-    pass
+import json
+from typing import Any
+from app.customers_shops.shops import Shop
+from app.customers_shops.customers import Customer
+
+
+def create_customers_shops() -> Any:
+    read_info = json.load(open("app/config.json"))
```
misha-doroshenko, line 8: Better to use a context manager

misha-doroshenko (changes requested):
You can also add more classes and methods to divide your project into simpler parts but it's optional

misha-doroshenko (approved):

app/customers_shops/customers.py
```diff
@@ -0,0 +1,65 @@
+from app.customers_shops.shops import Shop
+from dataclasses import dataclass
+import datetime
+
+
+@dataclass
+class Customer:
+    FUEL_PRICE: float
```
TPolina, line 8: it's not logical that the `Customer` class has a `fuel_price` attribute. Your class should be a model for real objects. Same with `shops_price`

TPolina: and why do you write class attributes with uppercase letters?

app/customers_shops/customers.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def calc_price(self, shop: Shop) -> (float, int):
+        self.shops_price[shop.name] = {
+            "price": 0,
+            "fuel": self.calc_fuel(shop),
+            "purchase": "",
+            "location": shop.location
+        }
+        for product, amount in self.product_cart.items():
+            self.shops_price[shop.name]["price"] +=\
```
TPolina, line 24: avoid using `\` to wrap line. Fix this everywhere

app/customers_shops/customers.py
```diff
@@ -0,0 +1,65 @@
...
+                shop.products[product] * amount
+            self.shops_price[shop.name]["purchase"] += (
+                f"{amount} {product}s "
+                f"for {shop.products[product] * amount} "
+                f"dollars\n")
```
TPolina, line 29: better style:
```
... += (
    ....
    ....
)
```
the close bracket should be on a separate line

TPolina (changes requested):

PR #110, opened 2022-10-11
Solution!

ketstap162:


app/main.py
```diff
@@ -1,3 +1,42 @@
...
-    pass
+import json
+
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        data = json.loads(file.read())
```
mltdll, line 9: There's a dedicated method for file reading in `json` module. Use `json.load(file)`.

app/car.py
```diff
@@ -0,0 +1,28 @@
...
+    @classmethod
+    def from_dict(cls, car: dict) -> Car:
+        return cls(brand=car["brand"],
+                   fuel_consumption=car["fuel_consumption"])
+
+    def fuel_cost(self,
+                  first_point: list[int],
+                  last_point: list[int],
+                  one_liter_price: float) -> int | float:
+        return round(
```
mltdll, line 21: I believe it's generally agreed that you should choose one style of line-splitting in your project. Therefore, I'd suggest to either reformat all other instances to this* format, or reformat this one to the other**. I would personally prefer the latter, but that's up to you.

*this:
```python
round(
    ...
)
```
**other: 
```python
round(...
      ...)
```

app/shop.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+import datetime
+from dataclasses import dataclass
+from math import ceil
+
+
+@dataclass
```
mltdll, line 8: It looks to me that this class has a little to much in the way of methods to be a dataclass. I might be wrong.

app/shop.py
```diff
@@ -0,0 +1,39 @@
...
+    def from_dict(cls, shop: dict) -> Shop:
+        return cls(name=shop["name"],
+                   products=shop["products"],
+                   location=shop["location"])
+
+    def is_in_stock(self, products: dict[str, int]) -> bool:
+        return all([product in self.products.keys()
+                    for product, count in products.items()])
+
+    def check(self,
```
mltdll, line 24: Check is not very good name in this case. I'd suggest to rename the method to something like `print_receipt`, and the variable to `receipt`.

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+                check = shop.check(person.name,
+                                   person.product_cart)
+                choice = {"name": shop.name,
+                          "check": check[1]}
+                cost = check[0]
+                cost += person.car.fuel_cost(person.location,
+                                             shop.location,
+                                             fuel_price)
+                if person.money >= cost:
+                    choices[cost] = choice
```
mltdll, line 31: What happens costs of trips to several stores have the same cost?

mltdll (changes requested):

app/main.py
ketstap162: Choose last of shops with same profit

mltdll (approved):

app/shop.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @classmethod
+    def from_dict(cls, shop: dict) -> Shop:
+        return cls(name=shop["name"],
+                   products=shop["products"],
+                   location=shop["location"])
+
+    def is_in_stock(self, products: dict[str, float]) -> bool:
+        return all([product in self.products.keys()
+                    for product, count in products.items()])
```
Nattalli, line 24: 👍🏻

Nattalli (approved):
Nice!

app/shop.py
mltdll: By the way @ketstap162, you can remove `[]` here. It would do the same thing, but faster and more memory-efficient.

Nattalli: yes, you are right 💯

PR #111, opened 2022-10-12
Solution

pazyn404:


tests/test_main.py
```diff
@@ -61,4 +61,3 @@ def test_shop_trip_output(monkeypatch):
 Monica doesn't have enough money to make purchase in any shop
 '''
     assert output == out
```
mltdll, line 63: It looks like you accidentally removed newline at the end of the file here. I'd suggest to put it back. :)

mltdll (changes requested):

tests/test_main.py
pazyn404: Did you mean that it should be:
```
Monica doesn't have enough money to make purchase in any shop

'''
``` ? )

Comment (mltdll) [reply tests/test_main.py:63]: No, i meant that it should be:
```python
assert output == out  # newline here

**Review by mltdll — APPROVED** [review]

File: app/representation/representation.py [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+        selected_shop = min(shops, key=lambda shop:
+                            customer.trip_cost(shop, self.fuel_price))
+
+        if customer.money < customer.trip_cost(selected_shop, self.fuel_price):
+            print(f"{customer.name} doesn't have enough money"
+                  f" to make purchase in any shop")
+            return
+
+        print(f"{customer.name} rides to {selected_shop.name}")
+        print()
```
Y-Havryliv, line 34: use `\n` instead of these prints (`print()`) everywhere

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #113, opened 2022-10-13
draft

hai-le-di:
Hey, i need help here. I ran this code on my computer and it works and I get the desired output, but when i run pytest i get AssertationError. Besides I don't know how to annotate instances of classes Shop and Customer in modules app.customer and app.shop.

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,0 +1,50 @@
+import math
+
+
+class Customer:
+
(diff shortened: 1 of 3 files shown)


IvanRamyk (changes requested):
please, add a type annotation to pass the flake8 tests

if you need a help feel free to ask in slack :)

PR #115, opened 2022-10-13
checklist added

DmytroSvirsa:


Abnormaltype (changes requested):
also, add:
make sure use write the correct path to the JSON file.
make sure u use correct import datetime
should be datetime.datetime.now()

Abnormaltype (approved):

PR #136, opened 2022-12-02
Solution

DevSpaciX:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        super().__init__(data_file)
+
+    def fuel_costs(self, shop: dict, customer: dict) -> None:
+        fuel_price = self.data["FUEL_PRICE"]
+        car_consumption = customer["car"]["fuel_consumption"]
+        distance = math.dist(customer["location"], shop["location"])
+        return round(distance * 2 / 100 * car_consumption * fuel_price, 2)
+
+    def how_much_costs_trip(self, customers: dict) -> dict:
+        customer_chart_milk = customers["product_cart"]["milk"]
```
OleksandrPozden, line 21: better say ```cost_of_the_trip``` or ```trip_cost``` 
Or even create a ```class Trip``` and then have a method or even a property as ```cost```

app/purchase.py
```diff
@@ -0,0 +1,37 @@
+class Purchase:
+    def __init__(self, data_file: dict) -> None:
+        self.data = data_file
+
+    def create_purchase(self, cheap_shop: dict, name: str) -> None:
+        print("Date: 04/01/2021 12:33:41")
```
OleksandrPozden, line 6: this can be as decorator. but it is not obligatory to do

app/purchase.py
```diff
@@ -0,0 +1,37 @@
...
+                    f"{bread_count} breads for "
+                    f"{product_to_buy['bread'] * bread_count} dollars"
+                )
+                total += product_to_buy["bread"] * bread_count
+                print(
+                    f"{butter_count} butters for "
+                    f"{product_to_buy['butter'] * butter_count} dollars"
+                )
+                total += product_to_buy["butter"] * butter_count
+                print(f"Total cost is {total} dollars")
```
OleksandrPozden, line 35: you repeat yourself here. Probably it is better to create a function that takes as an argument item that is bought and its cost and then prints it.

app/purchase.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def create_purchase(self, cheap_shop: dict, name: str) -> None:
+        print("Date: 04/01/2021 12:33:41")
+        chosen_shop = None
+        total = 0
+        for i in self.data["shops"]:
+            if i["name"] == cheap_shop:
+                chosen_shop = i
+        for item in self.data["customers"]:
+            milk_count = item["product_cart"]["milk"]
```
OleksandrPozden, line 13: By the way. You can think about customers as a class.

OleksandrPozden (changes requested):

app/purchase.py
DevSpaciX: Переробив )

app/Shop.py
```diff
@@ -0,0 +1,53 @@
+from app.Customer import Customer
+import math
+
+
+class Shop(Customer):
+    def __init__(
```
OleksandrPozden, line 6: You have done a lot, and I appreciate it but 
why Shop inherits Customers? They are two different entities

OleksandrPozden (changes requested):

app/Shop.py
DevSpaciX: Переробив !

app/Shop.py
```diff
@@ -0,0 +1,55 @@
...
+            cost_sum += self.get_product_cost(item) * customer.get_count(item)
+        return cost_sum
+
+    def info_buy(self, data: dict,
+                 cheap_shop: str,
+                 customer: Customer) -> None:
+        total = 0
+        for item in data:
+            if item["name"] == cheap_shop:
+                for i in item["products"]:
```
YuriiKindrat, line 49: Use more informable names of variables - not i

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+from app.Shop import Shop
+from app.Customer import Customer
+import json
+
+with open("app/config.json", "r") as data_customers:
+    data = json.load(data_customers)
+
+
+def shop_trip() -> None:
+    global shop
```
YuriiKindrat, line 10: what is "global shop" ?

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            print(f"Thanks, {customer.person_name}, for you purchase!")
+            print("You have bought: ")
+            shop.info_buy(data["shops"], cheap_shop[0], customer)
+            print(f"{customer.person_name} rides home")
+            print(
+                f"{customer.person_name} now has "
+                f"{customer.person_money - cheap_shop[1]} dollars\n"
+            )
+
+        cheap_shop = {}
```
YuriiKindrat, line 59: if u want to make this dict empty every iteration u can write it on line 12-13 after for loop

YuriiKindrat (changes requested):

app/main.py
DevSpaciX: Переробив) This task makes me mad -_-

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+                each_shop["name"],
+                each_shop["location"],
+                each_shop["products"],
+                item["name"],
+                item["product_cart"],
+                item["location"],
+                item["money"],
+                item["car"],
+                data["FUEL_PRICE"],
+            )
```
Abnormaltype, line 32: DRY, create a variables

tests/test_main.py
```diff
@@ -59,6 +59,5 @@ def test_shop_trip_output(monkeypatch):
 Monica's trip to the Shop '24/7' costs 16.84
 Monica's trip to the Central Shop costs 22.58
 Monica doesn't have enough money to make purchase in any shop
-'''
+"""
```
Abnormaltype, line 62: don't change the tests

Abnormaltype (changes requested):

Abnormaltype (changes requested):
check prev comments

YuriiKindrat (approved):

PR #146, opened 2022-12-15
Solution Maksym Maksimov

MKeSiMu:


app/customer.py
```diff
@@ -0,0 +1,75 @@
...
+            fuel_price: float
+    ) -> tuple:
+        total_cost = 0
+        min_cost = 100_000
+        cheapest_shop = ""
+        purchase_note = {"total": 0}
+        cheapest_purchase = {}
+
+        for shop in shop_list:
+            for key in customer.product_cart:
```
IvanRamyk, line 52: choose a more descriptive name for the loop variable

app/customer.py
```diff
@@ -0,0 +1,75 @@
...
+                car=Car(
+                    brand=customer["car"]["brand"],
+                    fuel_consumption=customer["car"]["fuel_consumption"]
+                )
+            )
+            customers_list.append(customer_object)
+        return customers_list
+
+    @staticmethod
+    def count_cost(
```
IvanRamyk, line 40: this function is really long. maybe it can be divided into several?

IvanRamyk (changes requested):

app/customer.py
katevasilenko: Still really long. Find some parts could be take out to another

katevasilenko (changes requested):

app/customer.py
```diff
@@ -0,0 +1,66 @@
...
+                customer.car.fuel_consumption,
+                fuel_price
+            )
+            total_cost = round(total_cost + trip_cost, 2)
+            print(f"{customer.name}'s trip to the {shop_option.name} "
+                  f"costs {total_cost}")
+
+            if total_cost < min_cost:
+                min_cost, cheapest_shop, cheapest_purchase = \
+                    total_cost, shop_option, purchase_note
```
Abnormaltype, line 65: put this logic into another method

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #153, opened 2022-12-30
Solution

sorrow03:


tests/test_main.py
```diff
@@ -1,5 +1,5 @@
 import datetime
-
+import pytest
```
olena-hul, line 2: Remove this change

app/customer.py
```diff
@@ -0,0 +1,22 @@
...
+
+@dataclasses.dataclass
+class Customer:
+    name: str
+    product_cart: dict
+    location: list
+    money: int
+    car: dict
+
+    def make_purchases(self, other: object) -> None:
```
olena-hul, line 13: Can you annotate `other` with its real type?

app/main.py
```diff
@@ -1,3 +1,38 @@
...
+        data = json.load(file)
+    fuel_price = data["FUEL_PRICE"]
+    customers = [
+        Customer(
+            customer["name"],
+            customer["product_cart"],
+            customer["location"],
+            customer["money"],
+            customer["car"])
+        for customer in data["customers"]]
```
olena-hul, line 17:
```suggestion
        for customer in data["customers"]
    ]
```

app/main.py
```diff
@@ -1,3 +1,38 @@
...
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        data = json.load(file)
+    fuel_price = data["FUEL_PRICE"]
+    customers = [
+        Customer(
+            customer["name"],
```
olena-hul, line 12: It's better practice to use keyword argumets when dataclasses are initialized. Using them, you are not needed to remember an order in which arguments should be passed

app/main.py
```diff
@@ -1,3 +1,38 @@
...
+        )
+        for shop in data["shops"]
+    ]
+    for customer in customers:
+        print(f"{customer.name} has {customer.money} dollars")
+        trips_price = sorted(
+            [shop.count_trip_price(customer, fuel_price) for shop in shops],
+            key=lambda x: x.trip_price
+        )
+        cheapest_shop = trips_price[0]
```
olena-hul, line 32: You can use `min()` to determine the cheapest shop. I think it will be more readable for this case

app/shop.py
```diff
@@ -0,0 +1,26 @@
...
+    def count_trip_price(self, other: object, fuel_price: float) -> object:
+        self.products_price = sum(
+            [
+                self.products[product] * count
+                for product, count in other.product_cart.items()
+            ]
+        )
+        distance = math.dist(self.location, other.location)
+        ride_price = 2 * fuel_price * (distance
+                                       * other.car["fuel_consumption"] / 100)
```
olena-hul, line 22:
```suggestion
        ride_price = 2 * fuel_price * (
            distance * other.car["fuel_consumption"] / 100
        )
```

app/shop.py
```diff
@@ -0,0 +1,26 @@
...
+        self.trip_price = round(ride_price + self.products_price, 2)
+        print(f"{other.name}'s trip to the "
+              f"{self.name} costs {self.trip_price}")
```
olena-hul, line 25: Not sure if this statement fits into 1 line, but if so - there is no need to split it into 2 lines

olena-hul (changes requested):

app/shop.py
olena-hul: Or make it in 1 line, if fits

app/customer.py
sorrow03: if i import class in this module, cycled import error will be

Abnormaltype (approved):
Well done!

PR #162, opened 2023-01-15
Solution

andriy-md:


app/location.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+from math import sqrt
+
+
+class Location:
+    def __init__(self, coordinates: list) -> None:
+        self.x = coordinates[0]
+        self.y = coordinates[1]
```
olena-hul, line 8: `self.x, self.y = coordinates`

app/customer.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            print(
+                f"{self.name} doesn't have enough "
+                f"money to make purchase in any shop"
+            )
+
+    def report_about_visit_to_shop(
+            self,
+            shop: Shop,
+            lowest_total_price: int | float) -> None:
```
olena-hul, line 71:
```suggestion
            lowest_total_price: int | float
        ) -> None:
```

app/customer.py
```diff
@@ -0,0 +1,90 @@
...
+        print(f"{self.name} rides to {shop.name}\n")
+        date_to_pass_test = datetime(2021, 1, 4, 12, 33, 41)
```
olena-hul, line 73: Why do you need this? You should use just datetime.today(), the mock is done in tests

app/customer.py
```diff
@@ -0,0 +1,90 @@
...
+            self.ride_to_shop_and_back_price(shop, fuel_price)
+            + self.products_in_a_shop_price(shop)
+        )
+        print(f"{self.name}'s trip to the {shop.name} costs {money_needed}")
+        return money_needed
+
+    def determine_cheapest_shop_and_visit_it(
+            self,
+            shops: list,
+            fuel_price: float) -> None:
```
olena-hul, line 51:
```suggestion
            fuel_price: float
        ) -> None:
```

olena-hul (changes requested):

app/customer.py
andriy-md: For some reason, if I put `datetime.now()`, I received an assertion error in PyCharm :(

app/car.py
```diff
@@ -0,0 +1,7 @@
+class Car:
+    def __init__(self, car: dict) -> None:
```
ArsenPidhoretskyi, line 2:
```suggestion
    def __init__(self, car: dict[str, Any]) -> None:
```

app/customer.py
```diff
@@ -0,0 +1,92 @@
...
+
+class Customer:
+    def __init__(self, customer: dict) -> None:
+        self.name = customer["name"]
+        self.products_to_buy = customer["product_cart"]
+        self.location = Location(customer["location"])
+        self.money = customer["money"]
+        self.car = Car(customer["car"])
+
+    def ride_to_shop_and_back_price(
```
ArsenPidhoretskyi, line 15: Rename to: get_trip_price

app/shop.py
```diff
@@ -0,0 +1,20 @@
...
+        for product in products_to_buy:
+            if product not in self.products:
+                return False
+        return True
+
+    def price_of_products(self, products_to_buy: dict) -> int | float:
+        total_price = 0
+        for product, quantity in products_to_buy.items():
+            total_price += self.products[product] * quantity
+        return total_price
```
ArsenPidhoretskyi, line 20: Use sum() function

ArsenPidhoretskyi (changes requested):
For function use verb names:
For example: money_needed_for_trip_and_products -> get_money_needed_for_trip_and_products

Also provide types for dict. For example dict -> dict[str, Any]

andriy-md:
Thanks for great comments

app/car.py
```diff
@@ -0,0 +1,10 @@
+class Car:
+    def __init__(self, car: dict[str, str | int | float]) -> None:
```
katevasilenko, line 2:
```suggestion
    def __init__(self, car: dict[str, Any]) -> None:
```

app/customer.py
```diff
@@ -0,0 +1,94 @@
+from typing import Any
+
+from datetime import datetime
```
katevasilenko, line 3:
```suggestion
from datetime import datetime
from typing import Any

```

app/customer.py
```diff
@@ -0,0 +1,94 @@
...
+from app.location import Location
+from app.car import Car
```
katevasilenko, line 5: Please be patient with order of imports. Car is alphabetically first

app/customer.py
```diff
@@ -0,0 +1,94 @@
...
+    def determine_cheapest_shop_and_visit_it(
+            self,
+            shops: list[Shop],
+            fuel_price: float
+    ) -> None:
+        print(f"{self.name} has {self.money} dollars")
+        lowest_total_price = float("inf")
+        best_shop = None
+        for shop in shops:
+            tmp = self.get_money_needed_for_trip_and_products(shop, fuel_price)
```
katevasilenko, line 59: It if a bad naming. I don’t get what tmp exactly means

app/customer.py
katevasilenko: Did you run your tests though terminal and command `pytest` Or using Run panel? Running though terminal should work with datetime.now()

katevasilenko (changes requested):

app/customer.py
andriy-md: I run through terminal and command `pytest` and get error (attached)

зображення

Vasyl-Smutok (approved):
Well Done 👍

PR #171, opened 2023-01-26
Solution

IhorPokr:


app/customer.py
```diff
@@ -0,0 +1,21 @@
...
+            money: int,
+            car: dict
+    ) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.money = money
+        self.car = car
+
+    def money_info(self) -> None:
```
ArsenPidhoretskyi, line 16: In this way use @property

app/customer.py
```diff
@@ -0,0 +1,21 @@
...
+        print(f"{self.name} has {self.money} dollars")
+
+    def less_money_info(self) -> None:
```
ArsenPidhoretskyi, line 19: In this way use @property

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        source_dict = json.load(shop_trip_file)
+
+    fuel_price = source_dict["FUEL_PRICE"]
+    customers = [Customer(
+        name=customer["name"],
+        product_cart=customer["product_cart"],
+        location=customer["location"],
+        money=customer["money"],
+        car=customer["car"])
+        for customer in source_dict["customers"]]
```
ArsenPidhoretskyi, line 18: Please format the code, so hard to read. In all places
```
[
    SomeClass(
        props
    )
    for props in something
]
```

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        name=shop["name"],
+        location=shop["location"],
+        products=shop["products"])
+        for shop in source_dict["shops"]]
+
+    for customer in customers:
+        car = Car(
+            brand=customer.car["brand"],
+            fuel_consumption=customer.car["fuel_consumption"]
+        )
```
ArsenPidhoretskyi, line 29: Store inside customer car object not dict

ArsenPidhoretskyi (changes requested):
Customer must have car as object not a dict. It will be better

app/customer.py
ArsenPidhoretskyi: Property for this function not for all atributes

ArsenPidhoretskyi (approved):
Looks good for me)

PR #173, opened 2023-01-31
Solution

yuliia-stopkyna:


app/customer.py
```diff
@@ -0,0 +1,85 @@
...
+            price * self.product_cart.get(product)
+            for product, price in shop.products_price.items()
+        ])
+        return round(
+            products_cost
+            + distance_to_shop * 2
+            * self.car.get_fuel_price_per_1_km, 2
+        )
+
+    def can_shop_here(self, shop: Shop) -> bool:
```
Abnormaltype, line 68: not clearly naming. If the function returns bool var, function should start with `is_`

app/shop.py
```diff
@@ -0,0 +1,46 @@
...
+            products: dict
+    ) -> None:
+        self.__name = name
+        self.__location = location
+        self.__products = products
+        Shop.shops[self.__name] = self
+
+    @property
+    def products_price(self) -> dict:
+        return self.__products
```
Abnormaltype, line 23: okay, if you want to use getters, I propose adding setters for the validation of incorrect data. WDYT?)

Abnormaltype (changes requested):

katevasilenko (approved):
Looks good to me

PR #174, opened 2023-02-03
Solution

aarrtemm:


app/main.py
```diff
@@ -1,3 +1,102 @@
...
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        data = json.load(file)
+
+    fuel_price = data["FUEL_PRICE"]
+    check_data = datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+
+    customers_list = [
```
ArsenPidhoretskyi, line 18: customers_list: customers is already mean not one)
customers_list -> customers
The same with shops_list, cars_list

app/main.py
```diff
@@ -1,3 +1,102 @@
...
+        for car in data["customers"]
+    ]
+
+    def product_cost(customer: Customer,
+                     shop: Shop,
+                     ) -> int | float:
+        milk_cost = customer.product_cart["milk"] * shop.products["milk"]
+        bread_cost = customer.product_cart["bread"] * shop.products["bread"]
+        butter_cost = customer.product_cart["butter"] * shop.products["butter"]
+        return milk_cost + bread_cost + butter_cost
```
ArsenPidhoretskyi, line 44: Move it to shop method

ArsenPidhoretskyi: Or to customer method

app/main.py
```diff
@@ -1,3 +1,102 @@
...
+
+    def cost_trip(customer: Customer,
+                  car: Car,
+                  shop: Shop
+                  ) -> float | int:
+        location = sqrt((shop.location[0] - customer.location[0]) ** 2
+                        + (shop.location[1] - customer.location[1]) ** 2)
+        cost_fuel = location * fuel_price * car.fuel_consumption / 100
+        return round(cost_fuel * 2, 2)
```
ArsenPidhoretskyi, line 53: The same. Move it to method of Shop or Customer

app/main.py
```diff
@@ -1,3 +1,102 @@
...
+
+    def go_to_home(remainder: int or float) -> str:
```
ArsenPidhoretskyi, line 55:
```suggestion
    def go_to_home(remainder: int | float) -> str:
```
Or Union[int, float]

app/main.py
```diff
@@ -1,3 +1,102 @@
...
+                          f"{products['butter'] * product_cart['butter']} "
+                          "dollars")
+                    total_cost = sum([value * shop.products[key]
+                                      for key, value in product_cart.items()])
+                    print(f"Total cost is {total_cost} dollars")
+                    print("See you again!\n")
+                    print(go_to_home(preferred_shop[0]))
+        else:
+            print(f"{customer.name} doesn't have enough "
+                  f"money to make purchase in any shop")
```
ArsenPidhoretskyi, line 98: Split to methods it is so hard to read)

ArsenPidhoretskyi (changes requested):

app/main.py
aarrtemm: do not quite understood

IvanRamyk (approved):

PR #180, opened 2023-02-07
'Solution'

timyr4iq:


app/car.py
```diff
@@ -0,0 +1,12 @@
...
+
+class Car:
+    def __init__(self, brand: str, fuel_consumption: float) -> None:
+        self.brand = brand
+        self.fuel_consumption = fuel_consumption
+
+    @classmethod
+    def make_instance(cls, car: dict) -> Car:
+        return cls(car["brand"],
+                   car["fuel_consumption"])
```
katevasilenko, line 12:
```suggestion
        return cls(
            car["brand"],
            car["fuel_consumption"]
        )
```
I would prefer this style with brackets all over the project. It would be great if you fix it

app/customer.py
```diff
@@ -0,0 +1,47 @@
...
+    def calculate_trip_costs(self,
+                             shops: List[Shop],
+                             fuel_price: float) -> str | Shop:
+        trip_costs = {}
+        for shop in shops:
+            trip_cost = 2 * (((shop.location[0] - self.location[0]) ** 2
+                              + (shop.location[1] - self.location[1]) ** 2)
+                             ** (1 / 2)) * (self.car.fuel_consumption
+                                            * fuel_price / 100)
+            for key in self.product_cart:
```
katevasilenko, line 37: Key is not good name for variable

app/main.py
```diff
@@ -1,3 +1,27 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+from app.customer import Customer
```
katevasilenko, line 2: Please add blank line between built-in and your own imports

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+    with open("app/config.json", "r") as file:
+        user_and_shop_data = json.load(file)
+    fuel_price = user_and_shop_data["FUEL_PRICE"]
+    customers = [Customer.make_instance(customer) for customer
+                 in user_and_shop_data["customers"]]
+    shops = [Shop.make_instance(shop) for shop in user_and_shop_data["shops"]]
+    is_first_customer = True
+    for customer in customers:
+        if not is_first_customer:
+            print("")
```
katevasilenko, line 16: Sorry, I don’t get this `is_first_customer` Could you explain plese?

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+        print(f"{customer.name} has {customer.money} dollars")
+        cheapest_shop = customer.calculate_trip_costs(shops, fuel_price)
+        if cheapest_shop != "not_enough_money":
+            print(f"{customer.name} rides to {cheapest_shop.name}\n")
+            cheapest_shop.print_bill(customer)
+            print(f"{customer.name} rides home")
+            print(f"{customer.name} now has {customer.money} dollars")
+        else:
+            print(f"{customer.name} doesn't have enough "
+                  f"money to make purchase in any shop")
```
katevasilenko, line 27: Move this to the separate function please

app/shop.py
```diff
@@ -0,0 +1,30 @@
...
+                   shop["location"],
+                   shop["products"])
+
+    def print_bill(self, customer: Callable) -> None:
+        current_time = datetime.datetime.now()
+        print("Date:", current_time.strftime("%d/%m/%Y %H:%M:%S"))
+        print(f"Thanks, {customer.name}, for you purchase!")
+        print("You have bought: ")
+        products_cost = 0
+        for key in customer.product_cart:
```
katevasilenko, line 24: Rename key variable please

katevasilenko (changes requested):

app/main.py
timyr4iq: I use this variable because there should be a blank line between the message about each user. If I print an empty line at the end of a message, then at the very end of the message I will have one extra empty line. so I check if the user is the first and in this case I don't print an empty line before him, but before all the rest I print. this way I avoid the extra line at the very end

Grayder0152 (approved):

PR #189, opened 2023-02-17
Solution

manjustice:


AndereLion (changes requested):

AndereLion (approved):

.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 inline-quotes = "
-ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
+ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818, F821
```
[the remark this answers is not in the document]
YuriiKindrat: no need to change this file

app/read_file_json.py
```diff
@@ -0,0 +1,14 @@
...
+
+def read_json() -> tuple:
+    with open("app/config.json", "r") as file:
+        data = load(file)
+
+    fuel_price = data["FUEL_PRICE"]
+    customers = create_customers(data["customers"])
+    shops = create_shops(data["shops"])
+
+    return fuel_price, customers, shops
```
YuriiKindrat, line 14: no need to create separate file for this func
also u can implement create_shops\create_customers func inside custormer and shop classes

app/create_instance.py
```diff
@@ -0,0 +1,41 @@
...
+
+from app.car import Car
+from app.customer import Customer
+from app.location import Location
+from app.product import Product
+from app.shop import Shop
+
+
+def create_customers(customers_dict: dict, ) -> List[Customer]:
+    customers = []
```
YuriiKindrat, line 11: u can create this list as class attribute using ur init method

app/create_instance.py
```diff
@@ -0,0 +1,41 @@
...
+                            location=location,
+                            money=customer["money"],
+                            car=car)
+        customers.append(customer)
+
+    return customers
+
+
+def create_shops(shops_dict: dict) -> List[Shop]:
+    shops = []
```
YuriiKindrat, line 28: the same here

YuriiKindrat (changes requested):

app/read_file_json.py
manjustice: I've talked to Yarik and he said I can add F821 to .flake8 because flake8 don't pass annotation as a literal

Serhii-Leonenko (changes requested):
All checks should be passed!

manjustice:
I know, but test_main was changed 3 days ago.

app/customer.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import annotations
```
wQuelS, line 1: Do you really need it? What python version are you on right now?)

wQuelS (approved):
Good job.
P.s. If you really need those `annotations` from `__future__` - please, think about updating python version)

PR #195, opened 2023-02-21
1 commit

AndereLion:


app/car.py
```diff
@@ -0,0 +1,17 @@
...
+        self.brand = brand
+        self.fuel_consumption = fuel_consumption
+
+    @staticmethod
+    def distance_calculator(
+            shop_location: list[int],
+            customer_location: list[int]
+    ) -> float:
+        return math.sqrt(
+            (shop_location[0] - customer_location[0])
```
bihunva, line 15: u can use 'math.dist' here

bihunva (changes requested):

app/car.py
AndereLion: Fixed

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+from app.logic_functions import start_trip
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file_in:
+        data = json.load(file_in)
+        fuel_price = data.get("FUEL_PRICE")
+        shops = Shop.shops_maker(data.get("shops"))
+        customers = Customer.customers_maker(data.get("customers"))
+        start_trip(customers, shops, fuel_price)
```
Serhii-Leonenko, line 16: Do not overload the context manager here, use it only for getting data and the other logic move outside the CM

AndereLion: Fixed !

wQuelS (approved):
Good job.
But think about updating your python version, to avoid using `from __future__ import annotations`

PR #198, opened 2023-02-22
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,0 +1,71 @@
+import json
+
+import math
+
+from dataclasses import dataclass
+
+import app.shop as mall
+from app.cart import Cart
+from app.car import Car
+
+
+@dataclass
+class Customer:
+    name: str
+    product_cart: list[Cart]
+    location: list[int]
+    money: float
+    car: Car
+
+    def calculate_trip_to_shop(self, shop: mall) -> float:
+        fuel_consumption = self.car.fuel_consumption
+        distance = self._count_distance(shop.location)
+        trip_cost = round(fuel_consumption * distance * 2
+                          / 100 * self._get_fuel_price(), 2)
+
+        return trip_cost
+
+    def print_report(self, shops: list[mall]) -> str:
+        shop_dict = {}
+        result = f"{self.name} has {self.money} dollars\n"
+
+        for shop in shops:
+            trip_sum = (self.calculate_trip_to_shop(shop)
+                        + shop.print_receipt(self)[1])
+            shop_dict[shop.name] = trip_sum
+
+            result += (
+                f"{self.name}'s trip to the {shop.name} costs {trip_sum}\n")
+
+        min_sum_trip = min(shop_dict.values())
+
+        if self.money >= min_sum_trip:
+
+            self.money -= min_sum_trip
+
+            prefer_shop = min(shop_dict, key=shop_dict.get)
+            result += f"{self.name} rides to {prefer_shop}\n\n"
+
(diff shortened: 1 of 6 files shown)


NEkropo1:
gj!

DavidGlusch (approved):

app/customer.py
```diff
@@ -0,0 +1,71 @@
+import json
+
+import math
+
+from dataclasses import dataclass
+
+import app.shop as mall
+from app.cart import Cart
+from app.car import Car
```
Serhii-Leonenko, line 9: Fix imports according to PEP8, also in other similar cases

app/customer.py
```diff
@@ -0,0 +1,71 @@
...
+        result = f"{self.name} has {self.money} dollars\n"
+
+        for shop in shops:
+            trip_sum = (self.calculate_trip_to_shop(shop)
+                        + shop.print_receipt(self)[1])
+            shop_dict[shop.name] = trip_sum
+
+            result += (
+                f"{self.name}'s trip to the {shop.name} costs {trip_sum}\n")
+
```
Serhii-Leonenko, line 39: Fix indentation here

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+from app.product import Product
+from app.car import Car
+from app.cart import Cart
+from app.shop import Shop
+from app.customer import Customer
+
+
+def shop_trip() -> None:
+    customers_obj = []
+    shops_obj = []
```
Serhii-Leonenko, line 12: Use more descriptive name

app/customer.py
```diff
@@ -0,0 +1,71 @@
...
+        return trip_cost
+
+    def print_report(self, shops: list[mall]) -> str:
+        shop_dict = {}
+        result = f"{self.name} has {self.money} dollars\n"
+
+        for shop in shops:
+            trip_sum = (self.calculate_trip_to_shop(shop)
+                        + shop.print_receipt(self)[1])
```
Serhii-Leonenko, line 34: On every iteration, you call the method calculate_trip_to_shop which calls on every iteration the _get_fuel_price() method which opens and reads the same data from the file on every iteration as well. It's not good, just get this data with only one opening the file.

Serhii-Leonenko (changes requested):

app/customer.py
SafonovVladimir: fixed

app/main.py
SafonovVladimir: Fixed

app/customer.py
SafonovVladimir: Done

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+        for product, price in shop.get("products").items():
+            shop_products.append(Product(product, price))
+
+        shops_list.append(Shop(shop_name, shop_location, shop_products))
+
+    for customer_obj in customers_list:
+        result += (customer_obj.print_report(shops_list)[:-2]
+                   if customer_obj == customers_list[-1]
+                   else customer_obj.print_report(shops_list))
```
YuriiKindrat, line 52: why do u need this loop, can u make this in the first loop? just move ur shop loop under the first loop

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: I've completely rewrote this function. 
Please check it.

YuriiKindrat (approved):

PR #201, opened 2023-02-23
Ready to work

NEkropo1:
I'll refactor according to readme and reupload.

### DIFF
--- a/app/client_resources/customer.py
+++ b/app/client_resources/customer.py
@@ -74,7 +74,7 @@ class Customer:
         if (min(shops_and_products_price.items(), key=lambda x: x[1])[1]
                 > self.money):
             print(f"{self.name} doesn't have enough money "
-                  f"to make purchase in any shop")
+                  f"to make a purchase in any shop")
             return
 
         best_place = min(shops_and_products_price.items(), key=lambda x: x[1])
--- a/app/client_resources/customer.py
+++ b/app/client_resources/customer.py
@@ -1,6 +1,7 @@
+import json
+
 from typing import List, Dict
 
-import json
 from app.shop_resources.shop import Shop
 from app.client_resources.Point2D import Point2d
 
--- a/app/client_resources/customer.py
+++ b/app/client_resources/customer.py
@@ -1,9 +1,13 @@
-import json
-
+from dataclasses import dataclass
 from typing import List, Dict
 
 from app.shop_resources.shop import Shop
 from app.client_resources.Point2D import Point2d
+
+
+@dataclass
+class ServerWithFuelData:
+    fuel_price: str
 
 
 class Customer:
@@ -26,15 +30,7 @@ class Customer:
         self._shops_price_and_map = None
 
     def check_fuel_price(self) -> None:
-        try:
-            with open("app/config.json", "r") as f:
-                world_data = json.load(f)
-        except FileNotFoundError as fnfe:
-            print("World didn't provide data to customer :(")
-            raise fnfe
-
-        self.fuel_price = world_data["FUEL_PRICE"]
-        del world_data
+        self.fuel_price = ServerWithFuelData.fuel_price
 
     def _count_price_and_map_to_shops(self) -> Dict[Shop, float]:
         self._shops_price_and_map = dict()
--- a/app/client_resources/customer.py
+++ b/app/client_resources/customer.py
@@ -83,4 +83,4 @@ class Customer:
         cheapest_shop = best_place[0]
         cheapest_shop.sell_products(self, self.product_cart)
         print(f"{self.name} rides home")
-        print(f"{self.name} now has {round(self.money, 2)} dollars")
+        print(f"{self.name} now has {round(self.money, 2)} dollars\n")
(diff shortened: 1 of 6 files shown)


MaybeTI (approved):

app/client_resources/customer.py
```diff
@@ -0,0 +1,86 @@
+from typing import List, Dict
+
+import json
+from app.shop_resources.shop import Shop
+from app.client_resources.Point2D import Point2d
```
Serhii-Leonenko, line 5: Fix imports according to PEP8 everywhere

app/main.py
```diff
@@ -1,3 +1,36 @@
...
+                     money=customer["money"],
+                     car=Car(**customer["car"]),
+                     known_shops=create_shops(world_data["shops"]))
+            for customer in world_data["customers"]]
+
+
+def shop_trip() -> None:
+    customers = process_customers_list()
+    for customer in customers:
+        customer.wake_up_and_ride()
```
Serhii-Leonenko, line 36: On every iteration, you call the method wake_up_and_ride which calls the method check_fuel_price which opens a file and gets the same data every time. It takes extra time, try to get all data with only one file opening.

Serhii-Leonenko (changes requested):

app/main.py
NEkropo1: Was pretty sure that program logic is to get exact fuel price by the time we getting the ride costs(I mean the data could be changed and that `json` is just hypothetical), so yup, it should've been checked every time when we're going to get a ride price... Idk which logic makes same price for every customer, on every ride. If it's like this, we can just simply make constants and get rid of file. I really can't get your idea. We _should_ check it every time, no?

app/client_resources/customer.py
```diff
@@ -0,0 +1,83 @@
...
+            print(
+                f"{self.name}'s trip to the {shop.name} "
+                f"costs {round(shops_and_products_price[shop], 2)}"
+            )
+
+        if (min(shops_and_products_price.items(), key=lambda x: x[1])[1]
+                > self.money):
+            print(f"{self.name} doesn't have enough money "
+                  f"to make a purchase in any shop")
+            return
```
YuriiKindrat, line 74: no need to use blank return here, u can just make backward condition

app/shop_resources/shop.py
```diff
@@ -0,0 +1,37 @@
...
+                 location: Point2d,
+                 products: dict[str, int]) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def sell_products(
+            self,
+            person: object,
+            products: dict[str]) -> None:
```
YuriiKindrat, line 17: fix indentation here

app/main.py
```diff
@@ -1,3 +1,39 @@
...
+                 shop_data["products"]) for shop_data in shops]
+
+
+def process_customers_list() -> List[Customer]:
+    try:
+        with open("app/config.json", "r") as f:
+            world_data = json.load(f)
+    except FileNotFoundError as fnfe:
+        print(f"Loading problem, {fnfe} doesn't exist.")
+        raise fnfe
```
YuriiKindrat, line 23: fnfe is bad name for exception, use full names for errors
no need to make print before raise exception

YuriiKindrat (changes requested):

app/client_resources/customer.py
NEkropo1: Yup, we can make just backward condition, but it will make code more complex. It's easy to understand, that if someone doesn't has enough money, program should print that message and pass/continue with next user,
Therefore I totally disagree with idea where we should change scary `blank return` for complex unreadable code. By complex and unreadable I mean, that program was working just human-like, in case, if there's not enough money - you don't ride to any shop, you don't even suggest and idea, that you should do anything else, But your suggestion is to make it more comples -> if you don't have not enough money, you make a ride, but if you don't have enough money, you don't ride, doesn't it sound complicated?

app/client_resources/customer.py
```diff
@@ -0,0 +1,90 @@
...
+                f"costs {round(shops_and_products_price[shop], 2)}"
+            )
+
+        if not self.could_manage_the_ride(shops_and_products_price):
+            print(f"{self.name} doesn't have enough money "
+                  f"to make a purchase in any shop")
+
+        if self.could_manage_the_ride(shops_and_products_price):
+            best_place = min(shops_and_products_price.items(),
+                             key=lambda x: x[1])
```
YuriiKindrat, line 83: gj, but why do u need to call this method twice ? u can just use else instead

app/client_resources/customer.py
YuriiKindrat: @NEkropo1 u are right about complicated solution, but there is no need to use scary `blank return` u can just use else keyword here

app/client_resources/customer.py
YuriiKindrat: so there is won't need to create a separate method, but it is up to u

YuriiKindrat (changes requested):

app/client_resources/customer.py
NEkropo1: Idk I just lost logic somewhere, sure else is better ))

YuriiKindrat (approved):

PR #203, opened 2023-02-23
Solution

WDemidenko:


app/main.py
```diff
@@ -1,3 +1,41 @@
...
-    pass
+from _datetime import datetime
+import json
+
+from app.trips_cost import calculate_trips_cost, Shop
+
+
+def shop_trip() -> None:
+    with open("config.json", "r") as file:
+        config = json.load(file)
```
amber-marichi, line 9: why dont you remove all this logic from context manager? i mean, it seems you just need to get the dict from json file into config variable and thats it, all other operations can work with stored data without open context manager.

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+            for shop in shops_config
+        ]
+        for customer in customers:
+            cheapest_cost, cheapest_shop, rest_of_money = calculate_trips_cost(
+                customer, shops, fuel_price
+            )
+            if rest_of_money >= 0:
+                print(f"{customer['name']} rides to {cheapest_shop[0]}")
+                date = datetime(2021, 4, 1, 12, 33, 41)
+                print(f"\nDate: {date.strftime('%m/%d/%Y %H:%M:%S')}\n"
```
amber-marichi, line 24: shouldnt the shop print the purchase check with current time according to the task? please check

amber-marichi (changes requested):
good job but i think you should really get al those operations out of the context manager :) another remark - i havnt noticed if you implemented that point from the task where the customers location should change to the shop location he arrives to.

app/main.py
WDemidenko: Unfortunately - not, to my surprise, the tests checked that the date would be a certain, from the readme

app/main.py
WDemidenko: Good point

amber-marichi (approved):
nice

YuriiKindrat (approved):

PR #210, opened 2023-02-27
Develop

amber-marichi:


ZabFTFT (approved):
Well done!

app/car.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __repr__(self) -> str:
+        return f"{self.brand} with fuel consumption {self.fuel_consumption}"
+
+    @classmethod
+    def car_from_dict(cls, car_info: dict) -> Car:
+        return cls(
+            brand=car_info.get("brand"),
+            fuel_consumption=car_info.get("fuel_consumption")
+        )
```
wQuelS, line 27: you could pass this `dict` right into `init` and initialize `brand` and `fuel_consumption`. So this method is a bit redundant

app/car.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
```
wQuelS, line 1: do you need it ?

app/customer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @classmethod
+    def customer_from_dict(cls, customer_info: dict) -> Customer:
+        return cls(
+            name=customer_info.get("name"),
+            money=customer_info.get("money"),
+            car=Car.car_from_dict(customer_info.get("car")),
+            location=customer_info.get("location"),
+            products=customer_info.get("product_cart"),
+        )
```
wQuelS, line 57: same. It could be done in `init`

app/shop.py
```diff
@@ -0,0 +1,59 @@
...
+    def __repr__(self) -> str:
+        return f"shop {self.name} located {self.location} has {self.products}"
+
+    @classmethod
+    def shop_from_dict(cls, shop_info: dict) -> Shop:
+        return cls(
+            name=shop_info.get("name"),
+            location=shop_info.get("location"),
+            products=shop_info.get("products")
+        )
```
wQuelS, line 34: same

wQuelS (changes requested):
Nice work! Left few comments below

app/car.py
amber-marichi: i see, i'll get them removed then. thanks!

app/car.py
amber-marichi: without classmethods probably not. thank you.

YuriiKindrat (approved):

PR #211, opened 2023-02-27
Solution

ZabFTFT:


app/main.py
```diff
@@ -1,3 +1,24 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+from app.customer import Customer
+from app.shop import Shop
```
DmytryTong, line 3: Add line between import type's please)

DmytryTong (approved):
Seems workable)

usrofgh (approved):

app/car.py
```diff
@@ -0,0 +1,34 @@
...
+    def calculates_trip_cost_to_shop(
+            customer: object,
+            shop: object
+    ) -> int | float:
+        trip_cost_to_shop = (
+            sqrt(
+                (shop.location[0] - customer.location[0]) ** 2
+                + (shop.location[1] - customer.location[1]) ** 2
+            )
+            * customer.car.calculates_fuel_cost_per_km()
```
wQuelS, line 32: you can use `dist` from `math`.

app/main.py
wQuelS, line 3: make your imports according to PEP8
https://peps.python.org/pep-0008/#imports

wQuelS (changes requested):

PR #212, opened 2023-02-27
Solution DonchykD

DmytryTong:


yura-hudzovskyi (approved):

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+                  f"costs {round(total_price, 2)}")
+
+        if len(ride_price) != 0:
+            shop_price_tuple = [(key, value)
+                                for key, value in ride_price.items()]
+            sorted_price_tuple = sorted(shop_price_tuple, key=lambda x: x[1])
+            sorted_price_dict = {key: value
+                                 for key, value in sorted_price_tuple}
+            shop_to_ride = next(iter(sorted_price_dict))
+
```
YuriiKindrat, line 53: simplify ur code here, why do u need to use ride_price as dict here? make a ride_price as list, and than u can sort it and return a first element

YuriiKindrat (changes requested):

PR #213, opened 2023-02-27
Solution

kuznetsb:


ada-krav (approved):

app/main.py
```diff
@@ -1,3 +1,37 @@
...
+            )
+        for shop in data["shops"]:
+            shops.append(
+                Shop(
+                    shop["name"],
+                    shop["location"],
+                    shop["products"]
+                )
+            )
+
```
Serhii-Leonenko, line 35: Do not overload the context manager here, use it only to get data from the file, and make the other logic outside it.

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #216, opened 2023-02-28
'Solution'

alexlega:


ZabFTFT (approved):
I tried.. xDDD

app/customer.py
```diff
@@ -0,0 +1,21 @@
+from app.car import Car
+from math import dist
```
Serhii-Leonenko, line 2: Fix imports according to PEP8

app/shop.py
```diff
@@ -0,0 +1,26 @@
...
+class Shop:
+    def __init__(self, shop: dict) -> None:
+        self.name = shop["name"]
+        self.products = shop["products"]
+        self.location = shop["location"]
+
+    def recipe(self, customer: Customer) -> None:
+        total_price = 0
+        print(
+            f"Date: 04/01/2021 12:33:41\nThanks, "
```
Serhii-Leonenko, line 20: Use datetime for this purpose

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            choose_shop.clear()
+            break
+        shop_to_ride.recipe(customer)
+
+        print(f"{customer.name} rides home")
+        print(f"{customer.name} now has "
+              f"{round((customer.money - min(choose_shop)), 2)} dollars\n")
+
+        ride_price_to_shop_list.clear()
+        choose_shop.clear()
```
Serhii-Leonenko, line 48: is this obligatory?

Serhii-Leonenko (changes requested):

app/main.py
alexlega: I use variables ride_price_to_shop_list and choose_shop for every iteration in the for loop and these variables were assigned before the "for" loop, so I need to clear these variables for every iteration. But I can assign them after "for" and then delete rows with the .clear() method

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+                  f"{shop.name} costs {total_price}")
+        shop_to_ride = shops[choose_shop.index(min(choose_shop))]
+        if customer.money > min(choose_shop):
+
+            print(f"{customer.name} rides to {shop_to_ride.name}\n")
+        else:
+            print(f"{customer.name} doesn't have "
+                  f"enough money to make purchase in any shop")
+            ride_price_to_shop_list.clear()
+            choose_shop.clear()
```
YuriiKindrat, line 39: what about to create this variables inside ur for loop?

app/main.py
```diff
@@ -1,3 +1,48 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+from app.customer import unpack_customers
+from app.shop import unpack_shops
+
+
+def shop_trip() -> None:
```
YuriiKindrat, line 6: simplify ur method shop_trip 
create more methods inside ur classes

YuriiKindrat (changes requested):
all tests should pass

PR #218, opened 2023-02-28
Solution

ada-krav:


app/shop.py
```diff
@@ -0,0 +1,17 @@
...
+        self.products = data["products"]
+
+    def calc_prod_price(self, customer_cart: dict) -> float:
+        total = 0
+        for item, count in customer_cart.items():
+            total += self.products[item] * count
+        return total
+
+    def calc_distance(self, customer_location: list) -> float:
+        x_delta = self.location[0] - customer_location[0]
```
Yulia-Mala, line 14: you can use math dist to make in one line but your code even better to understanding what is going on

Yulia-Mala (approved):
;)

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+                  f"to make purchase in any shop")
+            return None
+        the_shop = shops[totals.index(min(totals))]
+        print(f"{self.name} rides to {the_shop.name}")
+        return the_shop
+
+    def go_shopping(self, shop: Shop) -> None:
+        print()
+        # here should be datetime.now(), but tests are hard-coded
+        print("Date: 04/01/2021 12:33:41")
```
inrock1, line 44: I used strftime() method for datetime.now, and tests have been successfully done.

inrock1 (approved):

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+        if self.money < min(totals):
+            print(f"{self.name} doesn't have enough money "
+                  f"to make purchase in any shop")
+            return None
+        the_shop = shops[totals.index(min(totals))]
+        print(f"{self.name} rides to {the_shop.name}")
+        return the_shop
+
+    def go_shopping(self, shop: Shop) -> None:
+        print()
```
YuriiKindrat, line 42: it is better to use "\n" instead of blank print

YuriiKindrat (changes requested):

app/customer.py
ada-krav: I have just pushed commit with strftime 
As you can see check is failed

ada-krav: Screen from pycharm https://monosnap.com/file/fYfxUFreViDzqgBJNTdMphsg9TaBGC

PR #219, opened 2023-02-28
solution

usrofgh:


inrock1 (approved):

app/customer.py
```diff
@@ -0,0 +1,98 @@
+from datetime import datetime
+from math import dist
+from dataclasses import dataclass
+from app.car import Car
+from app.shop import Shop
+from app.car import FUEL_PRICE
```
wQuelS, line 6: imports should be according to PEP8. Blank lines, order, etc. Fix everywhere
https://peps.python.org/pep-0008/#imports

wQuelS (changes requested):

wQuelS (changes requested):
remove `config.json` from PR

YuriiKindrat (approved):

PR #225, opened 2023-03-01
Solution

Maytreyaya:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        for product, price in shop.get("products").items():
+            shop_products.append(Product(product, price))
+
+        shops_list.append(Shop(shop_name, shop_location, shop_products))
+
+    for customer in customers_list:
+        if customer == customers_list[-1]:
+            result += customer.print_message(shops_list)[:-2]
+        else:
+            result += customer.print_message(shops_list)
```
YuriiKindrat, line 52: why do u need 3 loops here, try to do it in one

YuriiKindrat (changes requested):

PR #230, opened 2023-03-06
Solution

ada-krav:


app/customer.py
```diff
@@ -0,0 +1,59 @@
+# from datetime import datetime
```
wQuelS, line 1: huh?

app/customer.py
```diff
@@ -0,0 +1,59 @@
+# from datetime import datetime
+from app.shop import Shop
+
+
+class Customer:
+    def __init__(self, info: dict) -> None:
+        self.name = info["name"]
+        self.location = info["location"]
+        self.money = info["money"]
+        self.car_consumption = info["car"]["fuel_consumption"]
```
wQuelS, line 10: why `car_fuel_consumption` is `Customer`'s instance attribute?
Maybe it is better to implement a separate `Car` class, what do you think?

app/customer.py
```diff
@@ -0,0 +1,59 @@
...
+                + 2
+                * fuel_price
+                * self.car_consumption
+                * shop.calc_distance(self.location)
+                / 100,
+                2,
+            )
+            print(f"{self.name}'s trip to the {shop.name} costs {trip_cost}")
+            totals.append(trip_cost)
+            self.trip_totals[shop.name] = trip_cost
```
wQuelS, line 32: Your `select_the_shop` breaks an SRP principle. This function has way too many responsibilities

app/customer.py
```diff
@@ -0,0 +1,59 @@
...
+            print(f"{self.name} doesn't have enough money "
+                  f"to make a purchase in any shop")
+            return None
+        the_shop = shops[totals.index(min(totals))]
+        print(f"{self.name} rides to {the_shop.name}")
+        return the_shop
+
+    def go_shopping(self, shop: Shop) -> None:
+        print("")
+        # print(f"Date: {datetime.now().strftime('%d/%m/%Y %H:%M:%S')}")
```
wQuelS, line 44: huh?

app/shop.py
```diff
@@ -0,0 +1,17 @@
...
+        total = 0
+        for item, count in customer_cart.items():
+            total += self.products[item] * count
+        return total
+
+    def calc_distance(self, customer_location: list) -> float:
+        x_delta = self.location[0] - customer_location[0]
+        y_delta = self.location[1] - customer_location[1]
+        distance = (x_delta**2 + y_delta**2) ** 0.5
+        return distance
```
wQuelS, line 17: you can use `dist` from `math` to do it more easily. Just google it

wQuelS (changes requested):

app/customer.py
YuriiKindrat: remove unnecessary print here

app/customer.py
```diff
@@ -0,0 +1,66 @@
...
+            item_total = shop.products[product] * quantity
+            print(f"{quantity} {product}s for {item_total} dollars")
+            total_cost += item_total
+        print(f"Total cost is {total_cost} dollars\n" f"See you again!\n")
+        self.money -= self.trip_totals[shop.name]
+        self.go_home()
+
+    def greeting(self) -> None:
+        print("")
+        # print(f"Date: {datetime.now().strftime('%d/%m/%Y %H:%M:%S')}")
```
YuriiKindrat, line 59: same

YuriiKindrat (approved):

PR #235, opened 2023-03-14
Develop

ZaikaDaria:
I have 1 issue with app/customers.py bill_by_shop(). In case there are enough money - everything works properly but when break works - the rest of the function continue its execution. I have tried to separate ill_by_shop() for 2 functions to avoid this exception but couldn't do it correctly. Will appreciate some help. Thanks

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,0 +1,36 @@
+from app.car import Car
+from app.shop import Shop
+
+
+class Customer:
+    def __init__(self, name, product_cart, location, money) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.money = money
+        self.car = Car
+
+    def has_money(self):
+        print(f"{self.name} has {self.money} dollars")
+
+    def total_shopping_list_price(self, shop: Shop) -> dict:
+        total_cost = {}
+        for product_name, price in shop.products.items():
+            total_cost[shop.name] = (price * self.product_cart.get(product_name, 0))
+        return total_cost
+
+    def price_summery(self, shops: Shop):
+        cheapest_shopping = []
+        for shop_name, total_amount in self.total_shopping_list_price(shops).items():
+            print(f"{self.name}'s trip to the {shop_name} costs {total_amount}")
+            cheapest_shopping.append(total_amount)
+            cheapest_shop = {i for i in self.total_shopping_list_price(shops) if
+                             self.total_shopping_list_price(shops)[i] == min(cheapest_shopping)}
+            if min(cheapest_shopping) <= self.has_money():
+                print(f"{self.name} rides to {cheapest_shop}")
+
+    def shopping(self, shops: Shop, fuel_cost: float) -> None:
+        print(f"{self.name} rides home")
+        print(
+            f"{self.name} now has {self.money - self.total_shopping_list_price(shops) - Car.calculate_road_expenses(shops, self.name, fuel_cost)} dollars\n")
+
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,36 +1,78 @@
+import datetime
+import math
+from typing import Any
+
 from app.car import Car
 from app.shop import Shop
 
 
 class Customer:
-    def __init__(self, name, product_cart, location, money) -> None:
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: list,
+            money: int,
+            car: Car
+    ) -> None:
+
         self.name = name
         self.product_cart = product_cart
         self.location = location
         self.money = money
-        self.car = Car
+        self.car = car
 
-    def has_money(self):
+    def has_money(self) -> None:
         print(f"{self.name} has {self.money} dollars")
 
-    def total_shopping_list_price(self, shop: Shop) -> dict:
-        total_cost = {}
-        for product_name, price in shop.products.items():
-            total_cost[shop.name] = (price * self.product_cart.get(product_name, 0))
-        return total_cost
+    def cost_per_km(self) -> float:
+        return self.car.fuel_consumption / 100
 
-    def price_summery(self, shops: Shop):
-        cheapest_shopping = []
-        for shop_name, total_amount in self.total_shopping_list_price(shops).items():
-            print(f"{self.name}'s trip to the {shop_name} costs {total_amount}")
-            cheapest_shopping.append(total_amount)
-            cheapest_shop = {i for i in self.total_shopping_list_price(shops) if
-                             self.total_shopping_list_price(shops)[i] == min(cheapest_shopping)}
-            if min(cheapest_shopping) <= self.has_money():
-                print(f"{self.name} rides to {cheapest_shop}")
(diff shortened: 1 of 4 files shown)


ZaikaDaria:
Знімок екрана 2023-03-15 о 01 01 11

app/customer.py
```diff
@@ -0,0 +1,78 @@
...
+from app.car import Car
+from app.shop import Shop
+
+
+class Customer:
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: list,
```
ArsenPidhoretskyi, line 14: list of what?

app/customer.py
```diff
@@ -0,0 +1,78 @@
...
+        self.money = money
+        self.car = car
+
+    def has_money(self) -> None:
+        print(f"{self.name} has {self.money} dollars")
+
+    def cost_per_km(self) -> float:
+        return self.car.fuel_consumption / 100
+
+    def calculate_road_expenses(self, shops: Any, fuel_cost: float) -> float:
```
ArsenPidhoretskyi, line 31: Shops is not Any

app/customer.py
```diff
@@ -0,0 +1,78 @@
...
+        for product, amount in self.product_cart.items():
+            price = (amount * shopping_price[product])
+            total_amount += price
+            print(f"{amount} {product}s for {price} dollars")
+
+    def product_cost(self, shops: Shop) -> float:
+        total_expanse = 0
+        for product, price in shops.products.items():
+            total_expanse += (price * self.product_cart.get(product))
+        return total_expanse
```
ArsenPidhoretskyi, line 46: Use sum instead

app/customer.py
```diff
@@ -0,0 +1,78 @@
...
+            if spent_for_shopping > self.money:
+                print(f"{self.name} doesn't have enough money "
+                      f"to make a purchase in any shop")
+                break
+
+            print(f"{self.name}'s trip to the {shop.name} "
+                  f"costs {spent_for_shopping}")
+            cheapest_shop[shop] = spent_for_shopping
+
+        chosen_shop = min(cheapest_shop, key=cheapest_shop.get)
```
ArsenPidhoretskyi, line 65: Check before if cheapest_shop is not empty

app/customer.py
```diff
@@ -0,0 +1,78 @@
...
+
+        for shop in shops:
+            spent_for_shopping = round((self.calculate_road_expenses(
+                shop, fuel_cost
+            )) * 2 + self.product_cost(shop), 2)
+
+            if spent_for_shopping > self.money:
+                print(f"{self.name} doesn't have enough money "
+                      f"to make a purchase in any shop")
+                break
```
ArsenPidhoretskyi, line 59: You need to do this after cycle when checking if cheapest_shop is not empty

ArsenPidhoretskyi (changes requested):

app/customer.py
ZaikaDaria: Please clarify why I should use sum() here? I didn't get it. Here I calculate the total expanse in shopping list that's why I multiply price and product amount. Thank you

ArsenPidhoretskyi (changes requested):

app/customer.py
ZaikaDaria: Got it. Thanks

ZaikaDaria: The situation is total_expanse hasn't been used in the function. Just calculated. I have deleted it at all. Thanks

ArsenPidhoretskyi: It used as I can see. Please use sum() function instead of custom calculation

ArsenPidhoretskyi (changes requested):

app/customer.py
ZaikaDaria: You are right. I had similar unused variable name in other cost_by_category() which i had deleted.

Regarding product_cost()
I am afraid I am stupid one.
When I use sum() the calculation goes wrong:

Знімок екрана 2023-03-16 о 13 44 42

Grayder0152: `total_expanse = sum([price * self.product_cart.get(product) for price, product_cart in shops.products.items()])`

Grayder0152 (changes requested):

app/customer.py
```diff
@@ -0,0 +1,80 @@
...
+        self.money = money
+        self.car = car
+
+    def has_money(self) -> None:
+        print(f"{self.name} has {self.money} dollars")
+
+    def cost_per_km(self) -> float:
+        return self.car.fuel_consumption / 100
+
+    def calculate_road_expenses(self, shops: Shop, fuel_cost: float) -> float:
```
katevasilenko, line 29: It’s better to call variable `shop` as only one shop expected on it

app/customer.py
```diff
@@ -0,0 +1,80 @@
...
+        distance = math.dist(self.location, shops.location)
+        return distance * self.cost_per_km() * fuel_cost
+
+    def cost_by_category(self, shopping_price: dict) -> None:
+        for product, amount in self.product_cart.items():
+            price = (amount * shopping_price[product])
+            print(f"{amount} {product}s for {price} dollars")
+
+    def product_cost(self, shops: Shop) -> float:
```
katevasilenko, line 38:
```suggestion
    def product_cost(self, shop: Shop) -> float:
```

app/customer.py
katevasilenko:
```suggestion
    def calculate_road_expenses(self, shop: Shop, fuel_cost: float) -> float:
```

app/customer.py
```diff
@@ -0,0 +1,80 @@
...
+        ) < self.money:
+            chosen_shop = min(cheapest_shop, key=cheapest_shop.get)
+            print(f"{self.name} rides to {chosen_shop.name}\n")
+
+            self.money -= float(cheapest_shop[chosen_shop])
+
+            data = datetime.datetime.now().strftime("%d/%m/20%y %H:%M:%S")
+            print(f"Date: {data}")
+            print(f"Thanks, {self.name}, for your purchase!")
+            print("You have bought: ")
```
katevasilenko, line 72: I believe you can write it in one print as you done it in 56-57 lines

app/customer.py
```diff
@@ -0,0 +1,80 @@
...
+            self.cost_by_category(chosen_shop.products)
+            print(f"Total cost is {self.product_cost(chosen_shop)} dollars")
+            print("See you again!\n")
+            print(f"{self.name} rides home")
+            print(f"{self.name} now has {round(self.money, 2)} dollars\n")
```
katevasilenko, line 77: The same remark here

app/main.py
```diff
@@ -1,3 +1,42 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+
+from app.car import Car
+from app.shop import Shop
+from app.customer import Customer
```
katevasilenko, line 5: Please sort imports

katevasilenko (changes requested):
Only few code style remarks, all other look good to me

Grayder0152 (approved):

PR #237, opened 2023-03-20
solution

viktory-koroliova:


Bohdan-Salamakha (changes requested):
fix tests

viktory-koroliova:
can u help? I don't get why tests are failed. it works okay in pycharm
image

app/shop.py
```diff
@@ -0,0 +1,5 @@
+class Shop:
+    def __init__(self, shop_info: dict) -> None:
+        self.name = shop_info["name"]
+        self.location = shop_info["location"]
+        self.products = shop_info["products"]
```
Bohdan-Salamakha, line 5: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

Bohdan-Salamakha (approved):

PR #238, opened 2023-03-23
Finish task

U-Ra-All:


app/customer.py
```diff
@@ -0,0 +1,82 @@
...
+                fuel_cost = (
+                    math.dist(self.location, shop.location)
+                    * self.car.fuel_consumption / 100
+                    * fuel_price
+                )
+
+                products_cost = 0
+
+                for key, value in self.product_cart.items():
+                    products_cost += value * shop.products.get(key)
```
ArsenPidhoretskyi, line 38: Use sum function

app/customer.py
```diff
@@ -0,0 +1,82 @@
...
+                trip_cost = round(fuel_cost * 2 + products_cost, 2)
+                print(f"{self.name}'s trip to the "
+                      f"{shop.name} costs {trip_cost}")
+
+                if not cheapest_trip_cost:
+                    cheapest_trip_cost = trip_cost
+                    selected_shop = shop
+                    continue
+
+                if trip_cost < cheapest_trip_cost:
```
ArsenPidhoretskyi, line 49: Union it in one if because code the same. Use or condition

ArsenPidhoretskyi (changes requested):

app/customer.py
U-Ra-All: Fixed

PR #242, opened 2023-03-25
Solution

yuraua5:


app/customer.py
```diff
@@ -0,0 +1,25 @@
...
+import math
+from app.shop import Shop
+from app.unpack_json_file import fuel_price
+
+
+@dataclasses.dataclass
+class Customer:
+    name: str
+    product_cart: dict
+    location: list
```
ArsenPidhoretskyi, line 11: list of what
provide inner type

app/customer.py
```diff
@@ -0,0 +1,25 @@
...
+        return round((math.dist(self.location, shop.location) * 2)
+                     * (self.car["fuel_consumption"] / 100) * fuel_price, 2)
+
+    def shop_total_cost(self, shop: Shop) -> list:
+        total_cost_list = []
+        for product, count in self.product_cart.items():
+            for shop_product, price in shop.products.items():
+                if product == shop_product:
+                    total_cost_list.append(count * price)
+        return total_cost_list
```
ArsenPidhoretskyi, line 25: It unoptimized
Use sum function and one loop
Also it must return float/int, not list

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+from app.customer import Customer
+from app.unpack_json_file import list_of_shops, list_of_customers
+
+
+def shop_trip() -> None:
+    for customer_data in list_of_customers:
+        customer = Customer(**customer_data)
+        print(f"{customer.name} has {customer.money} dollars")
+        choice_shop = ""
+        num = 1000000
```
ArsenPidhoretskyi, line 12: No it is wrong set to None and do check as
if num is None or customer.money < num:

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        print(f"{customer.name} rides to {choice_shop.name}\n\n"
+              f"Date: {current.strftime('%d/%m/%Y %H:%M:%S')}\n"
+              f"Thanks, {customer.name}, for your purchase!\n"
+              f"You have bought: ")
+        index = 0
+        price_choice_shop = customer.shop_total_cost(choice_shop)
+        for products, count in customer.product_cart.items():
+            print(f"{count} {products}s "
+                  f"for {price_choice_shop[index]} dollars")
+            index += 1
```
ArsenPidhoretskyi, line 36: It is not true also, because if order product_cart will be another. It will give wrong solution

app/shop.py
```diff
@@ -0,0 +1,8 @@
+import dataclasses
+
+
+@dataclasses.dataclass
+class Shop:
+    name: str
+    location: list
```
ArsenPidhoretskyi, line 7: list of what?

app/unpack_json_file.py
```diff
@@ -0,0 +1,7 @@
+import json
+
+with open("app/config.json", "r") as file:
+    data = json.load(file)
+    fuel_price = data["FUEL_PRICE"]
+    list_of_customers = data["customers"]
+    list_of_shops = data["shops"]
```
ArsenPidhoretskyi, line 7: Do this as function

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #249, opened 2023-04-10
Solution

anastasia-martyniuk:


EdAlekseiev (changes requested):
Tests have failed. Please, improve your solution

anastasia-martyniuk:
I had so many problems with this task, hope you won`t have any comments)

app/customer.py
```diff
@@ -0,0 +1,20 @@
+from app.car import Car
+
+
+class Customer:
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: [int],
```
olena-hul, line 9: For the 'location' parameter, use Tuple[int, int] instead of [int] to represent coordinates (x, y).

app/main.py
```diff
@@ -1,3 +1,127 @@
...
+
+        if min_price > customer.money:
+            print(
+                f"{customer.name} doesn't have enough money "
+                f"to make a purchase in any shop"
+            )
+        else:
+            print(f"{customer.name} rides to {shop_name}\n")
+            print(
+                f"Date: 04/01/2021 12:33:41\n"
```
olena-hul, line 105: Date and time are hardcoded in the output string. Consider using the `datetime` module to generate the current date and time.

app/main.py
```diff
@@ -1,3 +1,127 @@
...
-    pass
+import json
+import os.path
+
+from app.customer import Customer
+from app.shop import Shop
+from app.car import Car
+
+
+def price_of_gasoline(
```
olena-hul, line 9: Function name `price_of_gasoline` is not descriptive enough. Consider renaming it to `calculate_gasoline_cost` or similar to better reflect its purpose.

app/main.py
```diff
@@ -1,3 +1,127 @@
...
+        print(f"{customer.name} has {customer.money} dollars")
+        for shop in shop_list:
+            trip_price = price_of_gasoline(
+                customer, shop, fuel_price
+            ) + products_price(customer, shop)
+
+            if str(round(trip_price, 3))[-1] == "5":
+                trip_price = float(str(round(trip_price, 3))[:-1])
+            else:
+                trip_price = round(trip_price, 2)
```
olena-hul, line 85: If you need more accurate rounding, you can use Decimal module instead of doing such messy operations such as casting to string, etc.

olena-hul (changes requested):

app/shop.py
```diff
@@ -0,0 +1,16 @@
+from typing import Tuple
+
+
+class Shop:
+    def __init__(
+            self,
+            name: str,
+            location: Tuple[int, int],
+            products: dict
```
Bohdan-Salamakha, line 9: Consider using a more specific type for 'products' instead of 'dict'. You can use 'Dict[str, int]' where str represents the product name and int represents the quantity or price (depending on your implementation).

app/main.py
```diff
@@ -1,3 +1,128 @@
...
+            for shop in data["shops"]
+        ]
+        prices_dict = {}
+
+    for customer in customers_list:
+        print(f"{customer.name} has {customer.money} dollars")
+        for shop in shop_list:
+            trip_price = round(round(calculate_gasoline_cost(
+                customer, shop, fuel_price
+            ) + products_price(customer, shop), 3), 2)
```
Bohdan-Salamakha, line 83: There is an unnecessary nested `round()` function. Consider removing the inner `round()`.

app/main.py
```diff
@@ -1,3 +1,128 @@
...
+            print(
+                f"Total cost is "
+                f"{products_price(customer, shop_where_customer_go)} dollars\n"
+                f"See you again!\n"
+            )
+            print(
+                f"{customer.name} rides home\n"
+                f"{customer.name} now has "
+                f"{customer.money - min_price} dollars\n"
+            )
```
Bohdan-Salamakha, line 124: separate your function into a few functions with own logic

app/main.py
anastasia-martyniuk: i need both functions to round the numbers 28.215472 and 17.145232 to 28.21 and 17.14 if i use at once round(number, 2), i get 28.22 and 17.15.
math.floor() and math.ceil() also do not suit me, since other numbers should be rounded according to mathematical rules, so I found such a solution
the previous problem  solution was even worse:
            if str(round(trip_price, 3))[-1] == "5":
                trip_price = float(str(round(trip_price, 3))[:-1])
            else:
                trip_price = round(trip_price, 2)

ArsenPidhoretskyi (changes requested):
Good job!

PR #254, opened 2023-04-20
'Solution'

OleksandrYanchuk:


app/main.py
```diff
@@ -1,3 +1,11 @@
...
-    # write your code here
-    pass
+from app.trip.trip import Trip
+from app.customer.get_customers import get_list_of_customers
+from app.shop.get_shops import get_list_of_shops
+
+
+def shop_trip() -> None:
+    shops = get_list_of_shops()
+    customers = get_list_of_customers()
```
olena-hul, line 8: You're reading the file twice inside these functions. You can read the info from the file in advance and then pass the file content to the functions as an argument

olena-hul: And I'm not sure these functions should be in a separate module. You can include them here, in main

app/trip/trip.py
```diff
@@ -0,0 +1,103 @@
...
+        """Print summary of shopping session"""
+
+        milk_number = self.customer.product_cart.milk
+        milk_cost = milk_number * self.milk_price
+
+        bread_number = self.customer.product_cart.bread
+        bread_cost = bread_number * self.bread_price
+
+        butter_number = self.customer.product_cart.butter
+        butter_cost = butter_number * self.butter_price
```
olena-hul, line 87: I think you can use loop here

app/main.py
```diff
@@ -1,3 +1,11 @@
...
+
+    for customer in customers:
+        Trip(customer, shops)
```
olena-hul, line 11: You can initialize Trip class once above the loop

Then you can create go_shopping method and call it for every customer

app/trip/calculation.py
```diff
@@ -0,0 +1,33 @@
...
+def distance(customer: Customer, shop: Shop) -> float:
+    """Calculate distance between customer and shop."""
+    x1, y1 = customer.location
+    x2, y2 = shop.location
+    return math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
+
+
+def fuel_price() -> float:
+    """Return fuel_price"""
+    return get_data()["FUEL_PRICE"]
```
olena-hul, line 17: Again, redundant reading of file

app/trip/calculation.py
```diff
@@ -0,0 +1,33 @@
...
+    return round(
+        (fuel_consumption * distance_to_point * fuel_price() / 100) * 2, 2
+    )
+
+
+def shopping_cost(product_cart: ProductCart, price_list: PriceList) -> float:
+    """Calculate cost of all purchases in one shop"""
+    return (product_cart.milk * price_list.milk
+            + product_cart.butter * price_list.butter
+            + product_cart.bread * price_list.bread)
```
olena-hul, line 33: Loop can be used here

app/trip/trip.py
```diff
@@ -0,0 +1,103 @@
...
+        if self.customer_has_enough_money():
+            print(f"{self.customer.name} rides to {self.shop_to_ride}")
+            self.print_date()
+            self.print_summary()
+        else:
+            print(
+                f"{self.customer.name} doesn't have enough"
+                f" money to make a purchase in any shop"
+            )
+
```
olena-hul, line 30: A lot of extra logic in `__init__`, this should be moved to separate method

app/trip/trip.py
```diff
@@ -0,0 +1,103 @@
...
+    def customer_has_money(self) -> None:
+        """Print sum of money customer has."""
+        print(f"{self.customer.name} has {self.customer.money} dollars")
+
+    def trips_cost(self) -> None:
+        """Print the costs of customer`s shop trips."""
+        for shop in self.shops:
+            customer = self.customer
+            product_cart = customer.product_cart
+            price_list = shop.price_list
```
olena-hul, line 40: No need to use additional variables when you can just access instance attribute

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #255, opened 2023-04-20
'Solution'

HalynaPetrova:


app/calculate.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def trip_cost(customer: Customer, shop: Shop) -> float:
+    shop_trip_cost = fuel_cost(customer, shop)
+    shop_products_cost = products_cost(customer, shop)
+    total_trip_price = shop_trip_cost + sum(shop_products_cost.values())
+    return total_trip_price
+
+
+def ride_shop(customer: Customer, shops: list) -> dict:
```
ArsenPidhoretskyi, line 31: you can return just shop

app/calculate.py
```diff
@@ -0,0 +1,37 @@
+import operator
+from app.customer import Customer
+from app.shop import Shop
+from math import sqrt
+
+
+def fuel_cost(customer: Customer, shop: Shop) -> float:
+    fuel_price = 2.4
```
ArsenPidhoretskyi, line 8: No it comes from file

ArsenPidhoretskyi (changes requested):
customer.py does not exist

PR #263, opened 2023-05-02
Solution

bezhevets:


app/customer.py
```diff
@@ -0,0 +1,41 @@
...
+        self.location = location
+        self.money = money
+        self.car = car
+
+    def sum_products(self, product_shop: dict) -> float:
+        cost_products = 0
+        for name_product, quantity in self.product_cart.items():
+            if name_product in product_shop:
+                cost_products += quantity * product_shop[name_product]
+        return cost_products
```
ArsenPidhoretskyi, line 25: Use sum function

ArsenPidhoretskyi (changes requested):
Better to open file only once(do it inside shop_trip)

app/car.py
ArsenPidhoretskyi, line 1: This file is not needed

ArsenPidhoretskyi (approved):
Well done!

PR #265, opened 2023-05-04
Done

dirolius:


app/customer.py
```diff
@@ -0,0 +1,69 @@
...
+                     / 100) * fuel_price) * 2)
+
+        return fuel_cost
+
+    def get_trip_cost(self, fuel_price: float, shops: List[Shop]) -> dict:
+        dict_with_trip = {}
+        for shop in shops:
+            trip_cost = round(self.calculated_fuel_cost(fuel_price, shop)
+                              + self.get_product_cost(shop), 2)
+            dict_with_trip[shop.name] = trip_cost
```
ArsenPidhoretskyi, line 47: Maybe in this case better to save as key trip_cost and as value shop?

ArsenPidhoretskyi: It will simplify select_cheapest_trip function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,26 @@
...
+        ]
+        shops = [Shop(*shop.values()) for shop in data.get("shops")]
+        for customer in customers:
+            customer.start_balance()
+            dict_with_cost_trip = customer.get_trip_cost(fuel_price, shops)
+            cheapest_shop = customer.select_cheapest_trip(dict_with_cost_trip)
+            for shop in shops:
+                if shop.name == cheapest_shop:
+                    shop.print_receipt(customer)
+                    customer.ride_home_and_calculated_money()
```
ArsenPidhoretskyi, line 23: This is unlogical because you can return from select_cheapest_trip Shop. No need to return name
dict_with_cost_trip is key -> costs, value -> shop instance

app/customer.py
```diff
@@ -0,0 +1,66 @@
...
+                     / 100) * fuel_price) * 2)
+
+        return fuel_cost
+
+    def get_trip_cost(self, fuel_price: float, shops: List[Shop]) -> dict:
+        dict_with_trip = {}
+        for shop in shops:
+            trip_cost = round(self.calculated_fuel_cost(fuel_price, shop)
+                              + self.get_product_cost(shop), 2)
+            dict_with_trip[trip_cost] = shop.name
```
ArsenPidhoretskyi, line 47:
```suggestion
            dict_with_trip[trip_cost] = shop
```

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: It does not done Why you return from function shop.name
When you can return just shop?

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,26 @@
...
+        ]
+        shops = [Shop(*shop.values()) for shop in data.get("shops")]
+        for customer in customers:
+            customer.start_balance()
+            dict_with_cost_trip = customer.get_trip_cost(fuel_price, shops)
+            cheapest_shop = customer.select_cheapest_trip(dict_with_cost_trip)
+            for shop in shops:
+                if shop == cheapest_shop:
+                    shop.print_receipt(customer)
+                    customer.ride_home_and_calculated_money()
```
ArsenPidhoretskyi, line 23: Why you need for loop?
It is a little strange
```suggestion
            if cheapest_shop is not None:
                shop.print_receipt(customer)
                customer.ride_home_and_calculated_money()
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #277, opened 2023-05-18
solution

Dimdimior:


app/main.py
```diff
@@ -1,3 +1,34 @@
...
+    shops = []
+    for shop in data["shops"]:
+        shops.append(Shop(
+            shop["name"],
+            shop["location"],
+            shop["products"]
+        ))
+
+    for person in customers:
+        person.find_min_cost_shop(shops, fuel_price)
```
olena-hul, line 30: You're calling this method again in `trip` method. Choose only 1 place

olena-hul (changes requested):
Fix comments

app/shop.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass()
+class Shop:
+    name: str
+    location: list
```
Bohdan-Salamakha, line 7: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

PR #278, opened 2023-05-19
Done

MaksNochvai:


app/trip.py
```diff
@@ -0,0 +1,75 @@
...
+        print(
+            f"{count} {product if count < 2 else product + 's'} "
+            f"for {count * shop.provided_products.get(product, 0)} dollars"
+        )
+    print(f"Total cost is {round(total_cost, 2)} dollars")
+    print("See you again!")
+
+
+def go_to_shop(customer: Customer, shop: Shop, fuel_price: float) -> None:
+    trip_price = calculate_cost_of_the_trip(customer, fuel_price, shop)
```
olena-hul, line 70: You have already calculated trip_cost on a previous step

app/trip.py
```diff
@@ -0,0 +1,75 @@
...
+                fuel_price: float) -> int:
+    result = []
+    for shop in shops:
+        trip_costs = calculate_cost_of_the_trip(
+            customer, fuel_price, shop
+        )
+        result.append(trip_costs)
+        print(f"{customer.name}'s trip to the "
+              f"{shop.name} " f"costs {trip_costs}")
+    return result.index(min(result))
```
olena-hul, line 32: You can choose the min price directly in `for shop in shops` loop
No need to make redundant iterations
And you can return both the shop instance and the trip_costs

app/trip.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def can_afford_trip(customer: Customer,
+                    shops: list[Shop],
+                    fuel_price: float) -> None:
+    print(f"{customer.name} has {customer.money} dollars")
+    cheapest_shop = choose_shop(customer, shops, fuel_price)
+    trip_cost = calculate_cost_of_the_trip(
+        customer, fuel_price, shops[cheapest_shop]
+    )
```
olena-hul, line 42: You have already calculated cost_of_the_trip on a previous step

olena-hul (changes requested):

app/trip.py
YukihiroSM: Not fixed

YukihiroSM (changes requested):

PR #285, opened 2023-05-24
'Solution'

Evgen-Shepelenko:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+def shop_trip() -> None:
+
+    parent_dir = os.path.join(
+        "C:/",
+        "self_learning",
+        "Python",
+        "py-shop-trip",
+        "app"
+    )
```
ArsenPidhoretskyi, line 16: No it is not good. Because it will not work on my computer, because you say your path.
You can do:
parent_dir = os.path.dirname(__file__)

app/shop.py
```diff
@@ -0,0 +1,27 @@
...
+class Shop:
+    name: str
+    location: list
+    products: dict
+
+    def calculation_of_product_costs(self, product_cart: dict) -> float:
+        costs_of_products = 0.0
+        for product, quantity in product_cart.items():
+            costs_of_products += self.products[product] * quantity
+        return costs_of_products
```
ArsenPidhoretskyi, line 15: Use sum function

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        )for customer in config["customers"]
+    ]
+    shops = [
+        Shop(
+            shop["name"],
+            shop["location"],
+            shop["products"]
+        ) for shop in config["shops"]
+    ]
+    travel_shop = ""
```
ArsenPidhoretskyi, line 37:
```suggestion
    travel_shop = None
```

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    for customer in customers:
+        customer.print_amount_of_money()
+        min_cost_trip = 0
+        for shop in shops:
+            cost_trip = customer.cost_trip_for_the_products(shop, fuel_price)
+            if min_cost_trip == 0 or cost_trip < min_cost_trip:
```
ArsenPidhoretskyi, line 43:
```suggestion
            if min_cost_trip is None or cost_trip < min_cost_trip:
```

ArsenPidhoretskyi (changes requested):
Better to set defaults when searching for min or max to None

app/main.py
```diff
@@ -1,3 +1,46 @@
...
-    # write your code here
-    pass
+import json
+from os import path
+
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
```
vasylhnatiuk, line 8: Try to separate this function into a few functions.

vasylhnatiuk (changes requested):
you didn't do it  `When the customer arrives at the shop his location should equal to shop location.`

app/main.py
```diff
@@ -1,3 +1,51 @@
...
-    pass
+import json
+from os import path
+
+from app.customer import Customer
+from app.shop import Shop
+
+
+def loading_config_json() -> dict:
+    parent_dir = path.dirname("config.json")
```
ArsenPidhoretskyi, line 9:
```suggestion
    parent_dir = path.dirname(__file__)
```

ArsenPidhoretskyi: Whi will return path
/app at the end

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #289, opened 2023-05-29
Implement shop trip task

vitalii-babiienko:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        ]
+        shops = [
+            Shop(
+                shop["name"],
+                shop["location"],
+                shop["products"]
+            )
+            for shop in data["shops"]
+        ]
+        date = datetime(2021, 1, 4, 12, 33, 41).strftime("%d/%m/%Y %H:%M:%S")
```
ArsenPidhoretskyi, line 32: You can use datetime.datetime.now() instead of hardcoded value

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job

PR #290, opened 2023-05-30
Solution

markus2812:


tests/test_main.py
```diff
@@ -46,7 +46,7 @@ def test_shop_trip_output(monkeypatch):
 Thanks, Alex, for your purchase!
 You have bought: 
 2 milks for 4 dollars
-2 breads for 3.0 dollars
+2 breads for 3 dollars
```
vasylhnatiuk, line 49: why did you change it???

vasylhnatiuk: you can't change test data

app/utils.py
```diff
@@ -0,0 +1,103 @@
+import math
+import datetime
+from typing import Union
+from app.customer import Customer
+from app.shop import Shop
+
+
+def remove_trailing_zeros(num: float) -> float:
```
vasylhnatiuk, line 8: why do you need it?? you can use `round`

vasylhnatiuk (changes requested):

tests/test_main.py
markus2812: How can I get the desired result if in one case one result is displayed and in the other another.
image

app/utils.py
markus2812: Because depending on the result, I need different answers, and if I use round, the results are displayed with zero after the decimal point, and in most tests, such values are not allowed.

app/utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    cost += fuel_cost(customer, shop, fuel_price)
+
+    cost = round(cost, 2)
+    return cost
+
+
+def calculate_total_cost(customer: Customer, shop: Shop) -> None:
+    total_cost = 0
+    output = f"\nDate: " \
```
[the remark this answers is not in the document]
markus2812: How do I use ()? I'm just doing a simple transfer and PyCharm puts \

app/customer.py
```diff
@@ -0,0 +1,27 @@
...
+    location: list
+    home_location: list
+    money: int
+    car: dict
+
+
+def create_customer_from_file(data: dict) -> list[Customer]:
+
+    customers_data = data["customers"]
+
```
Grayder0152, line 17:
```suggestion
def create_customer_from_file(customers_data: dict) -> list[Customer]:
```

app/data.py
```diff
@@ -0,0 +1,12 @@
...
+from typing import Union
+from app.customer import create_customer_from_file
+from app.shop import create_shop_from_file
+
+
+def parse_config(file_path: str) -> Union[tuple]:
+    with open(file_path, "r") as file:
+        data = json.load(file)
+
+    return create_customer_from_file(data),\
```
Grayder0152, line 11: use `()` instead of `\`

app/main.py
```diff
@@ -1,3 +1,15 @@
-def shop_trip():
-    # write your code here
-    pass
+from app.utils import cheapest_shop, the_cost_of_shopping,\
+    calculate_total_cost, cash_balance
```
Grayder0152, line 2: use `()` instead of `\` and sort imports

app/shop.py
```diff
@@ -0,0 +1,19 @@
...
+@dataclasses.dataclass
+class Shop:
+    name: str
+    location: list
+    products: dict
+
+
+def create_shop_from_file(data: dict) -> list[Shop]:
+    shops_data = data["shops"]
+
```
Grayder0152, line 13:
```suggestion
def create_shop_from_file(shops_data: dict) -> list[Shop]:
```

Grayder0152 (changes requested):

vsmutok (approved):
Great 👍

PR #292, opened 2023-05-30
Solution

OksanaAdamchuk:


app/customer.py
```diff
@@ -0,0 +1,85 @@
...
+                (
+                    (
+                        shop.price_of_products_set(self.product_cart)
+                        + self.car.cost_of_way(fuel_price, distance) * 2
+                    )
+                ),
+                2,
+            )
+            print(f"{self.name}'s trip to the {shop.name} costs {total_cost}")
+            shops_costs[shop.name] = total_cost
```
ArsenPidhoretskyi, line 40: You can save as key total_cost and by value shop
And that you does not need the loop for minimal shop

app/shop.py
```diff
@@ -0,0 +1,14 @@
...
+    def __init__(self, name: str, location: Point, products: dict) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def price_of_products_set(self, products_list: dict) -> int | float:
+        products_value = 0
+        for key in products_list.keys():
+            products_value += self.products[key] * products_list[key]
+        return products_value
```
ArsenPidhoretskyi, line 14: use sum function

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #293, opened 2023-05-31
Had written `shop_trip` function with 4 modules containing classes wi…

N1khto:
…th data and logic for successful shopping trips.

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -2,18 +2,17 @@ from __future__ import annotations
 import datetime
 from dataclasses import dataclass
 from app.car import Car, Location
-from app.shop import Shop, Product
+from app.shop import Shop
 
 
 @dataclass
 class Customer:
     name: str
-    product_cart: list[Product]
+    product_cart: dict
     home_location: Location
     current_location: Location
     money: int | float
     car: Car
-    cheapest_shop: Shop = None
 
     def __repr__(self) -> str:
         return self.name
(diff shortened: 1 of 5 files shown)


app/customer.py
```diff
@@ -0,0 +1,83 @@
...
+
+@dataclass
+class Customer:
+    name: str
+    product_cart: list[Product]
+    home_location: Location
+    current_location: Location
+    money: int | float
+    car: Car
+    cheapest_shop: Shop = None
```
ArsenPidhoretskyi, line 16: No need to save cheapest_shop inside Customer

app/customer.py
```diff
@@ -0,0 +1,83 @@
...
+        fuel_cost = 2 * consumed_fuel * fuel_price
+        return round(fuel_cost, 2)
+
+    def calculate_products_price(self, shop: Shop) -> int | float:
+        total_cost = 0
+        for product_from_cart in self.product_cart:
+            for product_in_shop in shop.products:
+                if product_from_cart.name == product_in_shop.name:
+                    total_cost += (product_from_cart.quantity
+                                   * product_in_shop.price)
```
ArsenPidhoretskyi, line 36: Use sum function

app/customer.py
```diff
@@ -0,0 +1,83 @@
...
+            return
+
+    def shopping(self, shop: Shop) -> None:
+        self.current_location = shop.location
+        print(f"\nDate: "
+              f"{datetime.datetime.now().strftime('%d/%m/%Y %H:%M:%S')}")
+        print(f"Thanks, {self.name}, for your purchase!\nYou have bought: ")
+        for cart_product in self.product_cart:
+            for shop_product in shop.products:
+                if cart_product.name == shop_product.name:
```
ArsenPidhoretskyi, line 72: So complex loop. n^2
Better to save products as dict or use next + filter instead of for loop

app/customer.py
```diff
@@ -0,0 +1,83 @@
...
+                    print(f"{cart_product.quantity} {cart_product.name}s for "
+                          f"{(cart_product.quantity * shop_product.price)} "
+                          f"dollars")
+        print(f"Total cost is {self.calculate_products_price(shop)} dollars")
+        print("See you again!\n")
+
+    def go_home(self, shop: Shop, fuel_price: float) -> None:
+        self.current_location = self.home_location
+        print(f"{self.name} rides home")
+        self.money = self.money - self.calculate_trip_cost(shop, fuel_price)
```
ArsenPidhoretskyi, line 82:
```suggestion
        self.money -= self.calculate_trip_cost(shop, fuel_price)
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #299, opened 2023-06-09
Solution

vlamuha:


app/customer_class.py
```diff
@@ -0,0 +1,66 @@
...
+        cheapest_trip = min(trips)
+        if cheapest_trip < self.money:
+            self.money -= cheapest_trip
+            shop = trips[cheapest_trip]
+            print(f"{self.name} rides to {shop.name}\n")
+            return shop
+        else:
+            print(f"{self.name} doesn't have enough money "
+                  f"to make a purchase in any shop")
+            return None
```
ArsenPidhoretskyi, line 62: else and return None is not needed
```suggestion
        print(f"{self.name} doesn't have enough money "
              f"to make a purchase in any shop")
```

app/main.py
```diff
@@ -1,3 +1,25 @@
...
+from app.customer_class import Customer
+from app.shop_class import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        data = json.load(file)
+        fuel_price = data.get("FUEL_PRICE")
+        customers = [
+            Customer(*customer.values())
```
ArsenPidhoretskyi, line 12: if names the same you can use:
**customer

ArsenPidhoretskyi (approved):
Good job!

PR #304, opened 2023-06-15
Solution

yuro740:


app/shop.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Shop:
+
+    @classmethod
+    def from_dict(cls, shop_data: dict) -> list:
+        return [Shop(
+            shop["name"],
+            Point(shop["location"][0], shop["location"][1]),
+            shop["products"],
+        ) for shop in shop_data]
```
Grayder0152, line 13: Why classmethod? If you want to use classmethod you must use `cls` instead of `Shop`.
Or just use staticmethod

app/customer.py
```diff
@@ -0,0 +1,61 @@
...
+    @classmethod
+    def from_dict(cls, customers_data: dict) -> list:
+        return [Customer(
+                customer["name"],
+                customer["product_cart"],
+                Point(*customer["location"]),
+                customer["money"],
+                Car(customer["car"]["brand"],
+                    customer["car"]["fuel_consumption"]),
+                ) for customer in customers_data]
```
Grayder0152, line 19: Why classmethod? If you want to use classmethod you must use `cls` instead of `Shop`.
Or just use staticmethod

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #310, opened 2023-06-21
Wrong output, but I think in test wrong output :D

RedCatHot:


app/main.py
```diff
@@ -1,3 +1,39 @@
...
+    for customer in customers:
+        trip_costs = []
+        for shop in shops:
+            trip_cost = customer.calculate_trip_cost(shop, fuel_price)
+            trip_costs.append((shop, trip_cost))
+
+        trip_costs.sort(key=lambda x: x[1])
+
+        if trip_costs and customer.money >= trip_costs[0][1]:
+            print(f"\n{customer.name} has {customer.money} dollars")
```
diana-shyrokikh, line 24: When you just start to print information about the first customer you will have the empty line and only after that, for example, "Bob has 55 dollars". It's better to find out another way to put the "\n" which will not break the output about the first costumer

PR #323, opened 2023-07-05
Develop

vlodek69:


app/customer/customer.py
```diff
@@ -0,0 +1,81 @@
...
+    home_location: List[int]
+    product_cart: Dict[str, int]
+    money: int
+    car: Car
+    location: List[int] = None
+    chosen_shop: ChosenShop = None
+
+    def pick_shop(self, shops: List[Shop]) -> None:
+        print(f"{self.name} has {self.money} dollars")
+        self.chosen_shop = ChosenShop(self.money)
```
ArsenPidhoretskyi, line 23: Not true. Why do you init by self.money? Better to check inside code:

if self.chosen_shop is None or total_trip_cost <= self.chosen_shop.total_trip_cost:

app/customer/customer.py
```diff
@@ -0,0 +1,81 @@
...
+            one_way_trip_cost = (
+                self.car.fuel_consumption * 0.01 * distance_to_shop
+                * Car.fuel_price
+            )
+            shopping_cost = sum(
+                self.product_cart[product] * shop.products[product]
+                for product in self.product_cart
+            )
+
+            total_trip_cost = round(2 * one_way_trip_cost + shopping_cost, 2)
```
ArsenPidhoretskyi, line 41: Maybe better to move this to additional function or method

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #338, opened 2023-07-31
for checking

vkleshko:


app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+        total_cost = 0.0
+        date = datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+        print(f"Date: {date}")
+        print(f"Thanks, {self.name}, for your purchase!")
+        print("You have bought: ")
+
+        product_price_list = [value for value in shop.products.values()]
+        product_list = iter(product_price_list)
+
+        for key, value in self.product_cart.items():
```
ArsenPidhoretskyi, line 51: key, value is not clear names

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def time_for_shopping(self, shop: Shop) -> None:
+        total_cost = 0.0
+        date = datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+        print(f"Date: {date}")
+        print(f"Thanks, {self.name}, for your purchase!")
+        print("You have bought: ")
+
+        product_price_list = [value for value in shop.products.values()]
+        product_list = iter(product_price_list)
```
ArsenPidhoretskyi, line 49: It is a bad solution, if the order of product_cart and products will be different your code will not work correct

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+        best_shop = None
+
+        customer.initial_money()
+
+        for shop_data in shops_data:
+            shop_name = shop_data["name"]
+            shop_location = shop_data["location"]
+            products = shop_data["products"]
+
+            shop = Shop(shop_name, shop_location, products)
```
ArsenPidhoretskyi, line 39: Why you create a new shop for each customer?

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+                    customer_car.fuel_consumption,
+                    fuel_price
+                )
+            )
+
+            product_price_list = [value for value in products.values()]
+            product_list = iter(product_price_list)
+            for key, value in customer.product_cart.items():
+                cost = value * next(product_list)
+                total_cost += cost
```
ArsenPidhoretskyi, line 53: The same problem

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    shops_data = data["shops"]
+
+    shops = {}
+    for shop_data in shops_data:
+        shop_name = shop_data["name"]
+        shop_location = shop_data["location"]
+        products = shop_data["products"]
+
+        shop = Shop(shop_name, shop_location, products)
+        shops[shop_name] = shop
```
ArsenPidhoretskyi, line 23: FYI you can just save them as list

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+                customer.distance(
+                    shop.shop_location,
+                    customer_car.fuel_consumption,
+                    fuel_price
+                )
+            )
+
+            for product_name, price in customer.product_cart.items():
+                cost = price * shop.products[product_name]
+                total_cost += cost
```
ArsenPidhoretskyi, line 53: Use sum function

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi, line 53: Use sum function

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #340, opened 2023-08-02
Solution

pnakongit:


vsmutok (changes requested):
Please fix tests

app/shop.py
```diff
@@ -0,0 +1,46 @@
...
+    ) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def calculate_amount_for_products(self, product_cart: dict) -> int | float:
+        total_amount = 0
+        for product_name, count in product_cart.items():
+            total_amount += self.products.get(product_name, 0) * count
+        return total_amount
```
vsmutok, line 22: try to use `sum()` and  comprehension

vsmutok (changes requested):

app/shop.py
pnakongit: Thx a lot ! 
`sum()` and comprehension is a much better solution.

app/customer.py
```diff
@@ -0,0 +1,78 @@
...
+        self.car = Car(**car)
+        self.cheapest_shop = None
+        self.enough_money = False
+        self.total_cost = 0
+
+    def choose_the_best_shop(
+            self,
+            shops_list: List[Shop],
+            fuel_price: float
+    ) -> None:
```
ArsenPidhoretskyi, line 29: better to return from function best_shop. Because it to strange to save it inside the customer. And than pass shop into your next function

ArsenPidhoretskyi (changes requested):

app/customer.py
pnakongit: Thx you! Changed solution.

app/customer.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+
+    def buy_products(self, shop: Shop) -> None:
+        self.money -= shop.print_bill(
+            self.product_cart, self.name
+        )
+
+    def ride_to_home(self) -> None:
+        print(
+            f"{self.name} rides home\n"
```
vasylhnatiuk, line 77: change location to home image

vasylhnatiuk (changes requested):

app/customer.py
pnakongit: Done.

app/main.py
```diff
@@ -1,3 +1,39 @@
...
-    pass
+import json
+from pathlib import Path
+
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    file_path = Path(__file__).parent / "config.json"
```
Grayder0152, line 9: pls use os.path.join

Grayder0152 (changes requested):

app/main.py
pnakongit: Done!

ArsenPidhoretskyi (commented):
Well done!

ArsenPidhoretskyi (approved):

PR #359, opened 2023-08-29
Solution

nyanplague:


OlegatorLE (approved):
GJ!

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+        customer.ride_to_shop(cheapest_shop)
+
+        cheapest_shop.purchase(customer)
+
+        customer.ride_to_home(customer_home)
+
+        customer.count_money_after_shop(total_price)
+
+
+shop_trip()
```
IvanRamyk, line 42:
```
if __name__ == "__main__":
```

app/customer.py
```diff
@@ -0,0 +1,81 @@
...
+    def count_money_after_shop(self, total_cost: int) -> None:
+        self.money -= total_cost
+        print(f"{self.name} now has {round(self.money, 2)} dollars\n")
+
+
+def load_customers(file_data: json) -> list:
+    customers = []
+
+    for customer in file_data["customers"]:
+        customer = Customer(
```
IvanRamyk, line 69:
```suggestion
        customer = Customer.from_dict(customer)
```
makes sense to move initialization to classmethod

app/customer.py
```diff
@@ -0,0 +1,81 @@
...
+            customer["location"],
+            customer["money"],
+            Car(
+                customer["car"]["brand"],
+                customer["car"]["fuel_consumption"],
+            ),
+        )
+        customers.append(customer)
+
+    return customers
```
IvanRamyk, line 81: could be implemented using list comprehension

IvanRamyk (changes requested):

app/customer.py
nyanplague: @IvanRamyk A few tasks ago, mentors mentioned that these kind of methods might violate SOLID principles due to potential issues with inheritance. Is it acceptable in this case? :)

nyanplague: Still updated it , but let me know

nyanplague:
Tests pass locally, but not on GitHub. Everything I've changed relates only to fixes for the comments mentioned above. I don't know why they're not passing now - it's complaining about a float type of products.

IvanRamyk (approved):

PR #362, opened 2023-08-30
solution

innyshka:


RPodolets (approved):
It would be better to make classes more indepentend from config and put as arguments only necessary data

Barsh4ec (approved):

IvanRamyk (approved):

PR #370, opened 2023-08-31
Solution

9rosLove:


ORENYT (approved):
I don't think that using statickmethods a lot is a good practice, but well done as for me

PR #372, opened 2023-08-31
Implemented shop trip

KristianYe:


app/customer/car.py
```diff
@@ -0,0 +1,4 @@
+class Car:
+    def __init__(self, info: dict) -> None:
+        self.brand = info["brand"]
+        self.fuel_consumption = info["fuel_consumption"]
```
IvanRamyk, line 4:
```suggestion
    def __init__(self, brand, fuel_consumption) -> None:
        self.brand = brand
        self.fuel_consumption = fuel_consumption
    
    @classmethod    
    def from_dict(cls, car_dict: dict) -> Car:
        return cls(...)
```
it's a common way to create an alternative initializer

IvanRamyk: Also could be implemented as a dataclass

IvanRamyk (approved):
Could be refactored, but good job!

PR #389, opened 2023-09-19
solution

vsafian:


app/customers/customer.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class CustomerCar:
```
ArsenPidhoretskyi, line 5: Rename to Car

ArsenPidhoretskyi (approved):
Good job! FYI, try in future simplify your structure, for example class DickKeys is redundant better to use constants, day.py also is redundant

PR #402, opened 2023-10-04
solution

antonnech2309:


app/car.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass
+import math
```
Dimosphen1, line 2: `import` should precede `from ... import ...`

Dimosphen1 (approved):
LGTM, well done!

PR #412, opened 2023-10-17
Solution

sachavskyi:


app/customer.py
```diff
@@ -0,0 +1,67 @@
+import math
+import datetime
+from app.shop import Shop
```
viktoria-rybenchuk, line 3: Screenshot from 2023-10-13 13-38-45

fix imports according to PEP8

viktoria-rybenchuk (changes requested):

PR #414, opened 2023-10-24
Develop

mykolamateichuk:


Naz-iv:
final submit has failed git auto check, please fix

Naz-iv (changes requested):
test failed, please fix

app/classes/car.py
```diff
@@ -0,0 +1,17 @@
...
+                 fuel_consumption: int | float = 0.0) -> None:
+        self.brand = brand
+        self.fuel_consumption = fuel_consumption
+
+    @staticmethod
+    def from_dict(cls_dict: dict) -> Car:
+        new_car = Car()
+        for key in cls_dict:
+            setattr(new_car, key, cls_dict[key])
+        return new_car
```
LiudmylaKulinchenko, line 17: To create instances you can just use `**` to unpack and list comprehension. It will look much more readable, and you won't need to create FromDict class. Also think about method and parameter names to increase readability
And this method better to be classmethod, not static

LiudmylaKulinchenko: Same everywhere

app/classes/car.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import annotations
+from app.classes.from_dict import FromDict
+
+
+class Car(FromDict):
+    def __init__(self,
+                 brand: str = "",
+                 fuel_consumption: int | float = 0.0) -> None:
```
LiudmylaKulinchenko, line 8: Do you really want to accept creating cars with "" brand and 0.0 fuel_consumption?
Skip adding default values here, fix everywhere

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+
+def shop_trip() -> None:
+    config = util.parse_json("app/config.json")
+
+    fuel_price = config["FUEL_PRICE"]
+    customers = [Customer.from_dict(customer)
+                 for customer in config["customers"]]
+    shops = [Shop.from_dict(shop)
+             for shop in config["shops"]]
```
LiudmylaKulinchenko, line 13: You can just use unpacking, that I mentioned in previoius comments

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+                  f"doesn't have enough money to make a purchase in any shop")
+            continue
+
+        print(f"{customer.name} rides to {shops[closest_shop].name}\n")
+        shops[closest_shop].print_receipt(customer)
+        customer.money -= round(
+            util.calculate_trip_cost(fuel_price,
+                                     customer,
+                                     shops[closest_shop]),
+            2
```
LiudmylaKulinchenko, line 44: this `2` looks really lonely here (

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        )
+
+        print(f"{customer.name} rides home")
+        print(f"{customer.name} now has {customer.money} dollars\n")
```
LiudmylaKulinchenko, line 48: Consider split into methods

LiudmylaKulinchenko (changes requested):

PR #415, opened 2023-10-24
Solution

Naz-iv:


D1st3f (approved):
GJ!

app/instances/shop.py
```diff
@@ -0,0 +1,40 @@
...
+            f"Date: {datetime.datetime.now().strftime('%d/%m/%Y %H:%M:%S')}"
+            f"\nThanks, {customer.name}, for your purchase!\n"
+            "You have bought: \n"
+        )
+
+        for product, quantity in customer.product_cart.items():
+            cost = self.products[product] * quantity
+
+            if isinstance(cost, float) and cost % 1 == 0:
+                cost = int(cost)
```
LiudmylaKulinchenko, line 28: Why do you change it to integer here?

app/instances/shop.py
```diff
@@ -0,0 +1,40 @@
...
+            receipt += f"{quantity} {product}s for {cost} dollars\n"
+            total += cost
+
+        if total > 0:
+            receipt += f"Total cost is {total} dollars\nSee you again!\n"
```
LiudmylaKulinchenko, line 33: I don't think there will be a product that costs <= 0, so you can remove this check. What do you think?

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            print(f"{customer.name} doesn't have enough money to make "
+                  "a purchase in any shop")
+            continue
+
+        fuel_cost = fuel_cost_one_way(customer, fuel_price, shop)
+        customer.ride_to_shop(fuel_cost, shop.name, shop.location)
+
+        shop.serve_customer(customer)
+
+        customer.ride_home(fuel_cost)
```
LiudmylaKulinchenko, line 60: split into functions

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            Customer
+            (
+                customer["name"],
+                customer["product_cart"],
+                customer["location"],
+                customer["money"],
+                Car(customer["car"]["brand"],
+                    customer["car"]["fuel_consumption"])
+            )
+        )
```
LiudmylaKulinchenko, line 26: You can use `**` to unpack and list comprehenstion to create instances. Mind everywhere

LiudmylaKulinchenko (changes requested):

app/instances/shop.py
Naz-iv: sometimes cost is equal 3.0, so if it's a case (cost is float and decimal part is 0) I change cost to int to have 3 printed out instead of 3.0

app/instances/shop.py
Naz-iv: True, changed now

app/main.py
Naz-iv: when I try to unpack and do list comprehension (def init_customers(customers_data: dict) -> List[Customer]:
    return [
        Customer(**customer, car=Car(**customer["car"]))
        for customer in customers_data["customers"]
    ])
    I get error like this - TypeError: app.instances.customer.Customer() got multiple values for keyword argument 'car'
    Can you suggest how this can be fixed?

app/utils/utils.py
```diff
@@ -0,0 +1,97 @@
...
+            for product, quantity in customer.product_cart.items()), 2
+    )
+
+
+def trip_cost(customer: Customer, fuel_price: float, shop: Shop) -> float:
+    return round(fuel_cost_one_way(customer, fuel_price, shop) * 2
+                 + _products_cost(customer, shop), 2)
+
+
+def init_customers(customers_data: dict) -> List[Customer]:
```
LiudmylaKulinchenko, line 44: there is a difference between initialization and creating. You should change this and similar function names

app/utils/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    """ return [
+        Customer(**customer, car=Car(**customer["car"]))
+        for customer in customers_data["customers"]
+     ]"""
```
LiudmylaKulinchenko, line 48: It doesn't work because in customer you already have car. So when you pass `car` as a keyword argument, you pass it twice
You can fix that by creating car instance directly inside Customer init

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job ☀

PR #417, opened 2023-10-24
py-shop-trip Solution

RVChornyy:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        minimum_cost = float("inf")
+        for shop in shops_list:
+            shop_cost = (customer.calculate_trip_cost(shop, fuel_cost)
+                         + customer.shopping_cost(shop)["Total cost is "])
+            if shop_cost < minimum_cost:
+                minimum_cost = shop_cost
+                cheapest_shop = shop
+            print(f"{customer.name}'s trip to the "
+                  f"{shop.name} costs {shop_cost}")
+        if customer.money < minimum_cost:
```
prochigor, line 38: you should after if customer.money < minimum_cost: just contimue and than write without else

prochigor (changes requested):

app/main.py
RVChornyy: it is possible. byt why essential?

prochigor: mentors always reject it)

prochigor: not nested is better than nested is principle of python

PR #418, opened 2023-10-24
solution

kostomeister:


vshvanska (approved):

app/car.py
```diff
@@ -0,0 +1,13 @@
+class GetCarsData:
```
LiudmylaKulinchenko, line 1: You can create Car class. Don't forget about OOP

app/customer.py
```diff
@@ -0,0 +1,79 @@
+from app.shop import GetShopsData
+from app.car import GetCarsData
+
+
+class GetCustomersData:
```
LiudmylaKulinchenko, line 5: Same here. Your class should be equal to real object. Like Customer
GetCustomersData is not an some object. Fix that. OOP

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        )
+        if not isinstance(chosen_shop, GetShopsData):
+            print((f"{current_customer.name} doesn't have "
+                   f"enough money to make a purchase in any shop"))
+            continue
+        print(f"{current_customer.name} rides to {chosen_shop.name}")
+        print()
+        print("Date: 04/01/2021 12:33:41")
+        print(f"Thanks, {current_customer.name}, for your purchase!")
+        print("You have bought: ")
```
LiudmylaKulinchenko, line 33: You can use print once and `\n` inside a string

LiudmylaKulinchenko: Fix everywhere

app/shop.py
```diff
@@ -0,0 +1,21 @@
...
+    @property
+    def product_price(self) -> dict:
+        return self.shops_dict["products"]
+
+    @property
+    def location(self) -> list[int]:
+        return self.shops_dict["location"]
+
+    # def __lt__(self, other: ) -> GetShopsData:
+    #     return GetShopsData.location
```
LiudmylaKulinchenko, line 21: Remove that

LiudmylaKulinchenko (changes requested):

app/customer.py
```diff
@@ -0,0 +1,79 @@
+from app.shop import Shop
+from app.car import Car
+
+
+class Customer:
+    def __init__(self, customer_dict: dict) -> None:
+        self.customer_dict = customer_dict
+        self._money = self.customer_dict["money"]
```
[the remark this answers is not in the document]
LiudmylaKulinchenko: Then you could just use list comprehension and unpacking (using `**`) to create Customer and Shop instances
Hint: You can place Car creation inside Customer init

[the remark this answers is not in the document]
kostomeister: sorry didnt get what you meant at first

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    ]
+    fuel_price = all_data["FUEL_PRICE"]
+    for customer_data in customers_data:
+        current_customer = Customer(
+            customer_data["name"],
+            customer_data["money"],
+            customer_data["product_cart"],
+            customer_data["location"],
+            customer_data["car"],
+        )
```
LiudmylaKulinchenko, line 25: Same here. List comprehension and create customer like this `Customer(**customer_data)`
Using ** like this calles unpacking

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if not isinstance(chosen_shop, Shop):
+            print(
+                f"{current_customer.name} doesn't have enough money"
+                f" to make a purchase in any shop"
+            )
+            continue
+        print(f"{current_customer.name} rides to {chosen_shop.name}\n")
+        print("Date: 04/01/2021 12:33:41")
+        print(f"Thanks, {current_customer.name}, for your purchase!")
+        print("You have bought: ")
```
LiudmylaKulinchenko, line 39: Use one print, you can make a new line using `\n` inside a string. Fix everywhere

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+        chosen_shop = current_customer.choose_shop(shops_list, fuel_price)
+        if not isinstance(chosen_shop, Shop):
+            print(
+                f"{current_customer.name} doesn't have enough money"
+                f" to make a purchase in any shop"
+            )
+            continue
+        print(f"{current_customer.name} rides to {chosen_shop.name}\n")
+        print("Date: 04/01/2021 12:33:41")
```
LiudmylaKulinchenko, line 37: Use datetime.now() instead

LiudmylaKulinchenko (changes requested):

app/main.py
kostomeister: image
its not working with datetime.now()

kostomeister: the problem is in tests, its not working with ```from datetime import datetime``` but its working with ```import datetime```

LiudmylaKulinchenko (approved):
Well done! ☀

PR #422, opened 2023-10-25
Solution

vshvanska:


app/customers.py
```diff
@@ -0,0 +1,55 @@
+from app.cars import Car
+from app.shops import Shop
+
+
+class Customer:
+    customers = []
```
LiudmylaKulinchenko, line 6: No need to store them here

app/shops.py
```diff
@@ -0,0 +1,28 @@
+class Shop:
+    shops = []
```
LiudmylaKulinchenko, line 2: same here

app/shops.py
```diff
@@ -0,0 +1,28 @@
...
+        self.products = products
+        Shop.shops.append(self)
+
+    def __repr__(self) -> str:
+        return self.name
+
+    def print_bill(self, customer_name: str, product_cart: dict) -> None:
+        print("Date: 04/01/2021 12:33:41")
+        print(f"Thanks, {customer_name}, for your purchase!")
+        print("You have bought: ")
```
LiudmylaKulinchenko, line 19: You can use pring once and `\n` inside a string

app/dataloader.py
```diff
@@ -0,0 +1,38 @@
...
+        Car.fuel_price = self.data["FUEL_PRICE"]
+
+    def create_customers_and_cars(self) -> None:
+        for customer in self.data["customers"]:
+            Customer(customer["name"],
+                     customer["product_cart"],
+                     customer["location"],
+                     customer["money"],
+                     Car(customer["car"]["brand"],
+                         customer["car"]["fuel_consumption"]))
```
LiudmylaKulinchenko, line 32: Use list comprehension and unpacking (using `**`)
Hint: You can place Car creation inside Customer init

LiudmylaKulinchenko (changes requested):

app/cars.py
```diff
@@ -0,0 +1,16 @@
+from math import sqrt
+
+
+class Car:
+    fuel_price = 0
```
LiudmylaKulinchenko, line 5: There is no need to store it as a class attribute here. Fuel price depends on where are you going. You don't need to store this value outside or a function or method where you use it

app/dataloader.py
```diff
@@ -0,0 +1,34 @@
...
+    ]
+
+
+def create_shops(data: dict) -> list:
+    return [
+        Shop(shop["name"],
+             shop["location"],
+             shop["products"])
+        for shop in data["shops"]
+    ]
```
LiudmylaKulinchenko, line 34: You missed my previous comment
Use list comprehension and unpacking (using `**`)
Same in Customer and Shop creation

app/shops.py
LiudmylaKulinchenko: You missed my comment also here)

LiudmylaKulinchenko (changes requested):
Please, be considerate and make corrections accordingly to all comments

app/cars.py
LiudmylaKulinchenko: You didn't fix that

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! ☀

PR #430, opened 2023-10-28
Solution

petrykivd:


sakhaline (approved):

app/customer.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @classmethod
+    def create_customers(cls, customers: List) -> List:
+        customer_objects = []
+        for customer in customers:
+            car_data = customer.pop("car")
+            car = Car(**car_data)
+            customer["car"] = car
+            customer_objects.append(cls(**customer))
+        return customer_objects
```
LiudmylaKulinchenko, line 24: U can place Car creation inside Customer init. Then create customer just using unpacking directly where you need that and remove this method
It will pretty simplify your code

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+    fuel_price = config.get("FUEL_PRICE")
+    customers = [
+        Customer(
+            name=customer["name"],
+            product_cart=customer["product_cart"],
+            location=customer["location"],
+            money=customer["money"],
+            car_data=customer["car"]
+        ) for customer in config.get("customers")
+    ]
```
LiudmylaKulinchenko, line 23: You forgot to use unpacking here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! ☀

PR #431, opened 2023-10-28
Solution

sakhaline:


OlehShumov (approved):

app/customer.py
```diff
@@ -0,0 +1,74 @@
+from math import dist
+from typing import Dict, Any
+
+from app.car import Car
+
+
+class Customer:
+    def __init__(self, customer_data: Dict[str, Any]) -> None:
```
LiudmylaKulinchenko, line 8: When you create Customer object, you should every time look here through all the attributes to be sure that u put right customer_data dict here. Don't do that, make it easier

app/customer.py
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return self.__possible_shops
+
+    def choose_cheapest_shop(self) -> dict:
+        cheapest_shop_data = min(
+            self.__possible_shops.items(), key=lambda item: item[1]
+        )
+        self.__cheapest_shop = {
+            "shop": cheapest_shop_data[0], "total_cost": cheapest_shop_data[1]
+        }
```
LiudmylaKulinchenko, line 61: Try to rewrite to make more readable. Don't use lambda like this, it is not clear code
Maybe use unpacking for cheapest_shop_data

app/json_parser.py
```diff
@@ -0,0 +1,23 @@
...
+from typing import Any
+
+
+def read_json_from_file(file_path: str) -> Any:
+    try:
+        with open(file_path, "r") as file:
+            json_data = json.load(file)
+        return json_data
+    except FileNotFoundError:
+        raise FileNotFoundError(f"File not fount: {file_path}")
```
LiudmylaKulinchenko, line 11: Don't catch this error here, you just raise it again with defferent error message. No need to do that

app/json_parser.py
```diff
@@ -0,0 +1,23 @@
...
+def get_customers_data(json_data: Any) -> list:
+    return json_data.get("customers")
+
+
+def get_shops_data(json_data: Any) -> list:
+    return json_data.get("shops")
+
+
+def get_fuel_price(json_data: Any) -> float:
+    return json_data.get("FUEL_PRICE")
```
LiudmylaKulinchenko, line 23: If u use `get` here then u are not sure that function will return those types. They also could be None
And you don't really need to create function for so simple actions

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+def shop_trip() -> None:
+    data = app.json_parser.read_json_from_file("app/config.json")
+    fuel_price = app.json_parser.get_fuel_price(data)
+
+    customers = [
+        Customer(customer_data)
+        for customer_data
+        in app.json_parser.get_customers_data(data)
+    ]
+    shops = [
```
LiudmylaKulinchenko, line 15: Rewrite using unpacking `**customer`
You can place Car creation inside `Customer.__init__`

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        Shop(shop_data)
+        for shop_data
+        in app.json_parser.get_shops_data(data)
+    ]
```
LiudmylaKulinchenko, line 19: rewrite using unpacking

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+                  f"costs {possible_shops[shop]}")
+
+        cheapest_shop = customer.choose_cheapest_shop()
+
+        if customer.evaluate_shopping_possibility():
+            customer.go_to_the_shop()
+            print(f"{customer.name} rides to {cheapest_shop['shop'].name}"
+                  "\n")
+
+            cheapest_shop["shop"].print_purchase_receipt(customer)
```
LiudmylaKulinchenko, line 39: create variable with cheapest_shop to not call it twice

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            shops, fuel_price
+        )
+
+        for shop in shops:
+            print(f"{customer.name}'s trip to the {shop.name} "
+                  f"costs {possible_shops[shop]}")
+
+        cheapest_shop = customer.choose_cheapest_shop()
+
+        if customer.evaluate_shopping_possibility():
```
LiudmylaKulinchenko, line 34: Put your condition here to make logic more clear

app/shop.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def serve_customer(self, customer: Customer) -> Dict[str, dict]:
+        sold_products = {}
+
+        for product in customer.product_cart:
+            amount = customer.product_cart[product]
+            cost = customer.product_cart[product] * self.products[product]
+
+            if cost == int(cost):
+                cost = int(cost)
```
LiudmylaKulinchenko, line 21: You can make it in one line

LiudmylaKulinchenko (changes requested):

app/customer.py
sakhaline: I like, the way it looks with lambda, if you don't mind, I'll remain it as it is. ^‿~

LiudmylaKulinchenko (approved):
Good job 😸

PR #432, opened 2023-10-29
Solution

OlehShumov:


SelFisher1488 (approved):
nice code! but can't understand why you using datetime.now and it works, when it is not working for me :(

LiudmylaKulinchenko (changes requested):

PR #435, opened 2023-10-29
sol

oleksandr-pupchenko:


fedorov-s-od (approved):
Probably, in the nearest future, we will write more concise code. At this stage, working code is a good code ;) Good job! )

app/customer.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @classmethod
+    def create_customers(cls, customers: List) -> List:
+        customer_objects = []
+        for customer in customers:
+            car_data = customer.pop("car")
+            car = Car(**car_data)
+            customer["car"] = car
+            customer_objects.append(cls(**customer))
+        return customer_objects
```
LiudmylaKulinchenko, line 24: You can place `Car(**car)` inside Customer init and just use unpacking for Customer creation

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (changes requested):
All checks should pass

LiudmylaKulinchenko (approved):
Well done! 🚀

PR #440, opened 2023-11-04
solution

spryima:


app/data_configuration.py
```diff
@@ -0,0 +1,49 @@
...
+            customer["product_cart"],
+            customer["location"],
+            customer["money"],
+            Car(
+                customer["car"]["brand"],
+                customer["car"]["fuel_consumption"],
+            )
+        ) for customer in big_dict["customers"]]
+
+    def create_shops_obj(self, big_dict: dict) -> None:
```
ArsenPidhoretskyi, line 44: In this case it must be named load_shops(), the same with customers

app/data_configuration.py
```diff
@@ -0,0 +1,49 @@
+import json
+import os
+
+from app.car import Car
+from app.shop import Shop
+from app.customer import Customer
+
+
+class Data:
```
ArsenPidhoretskyi, line 9: Rename Data to Configuration

app/shop.py
```diff
@@ -0,0 +1,35 @@
...
+                 location: list[int],
+                 products: dict[str, float],
+                 ) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def check_prices(self, prod_cart: dict[str, int]) -> float:
+        return sum(self.products[prod] * qty
+                   for prod, qty in prod_cart.items())
```
ArsenPidhoretskyi, line 16: No need to name variables as abbreviation rename qty to quantity

ArsenPidhoretskyi (changes requested):

Yurnerroo (approved):

PR #441, opened 2023-11-07
Solution

loplicat:


Polyakiv-Andrey (changes requested):
Tests not passed!

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        trips_prices = []
+        for shop in shops:
+
+            fuel_expenses = customer.car.calculate_fuel_expenses(
+                customer.calculate_distance(shop.location),
+                fuel_price
+            )
+            trip_cost = shop.calculate_trip_cost(
+                shop.calculate_product_cost(customer),
+                fuel_expenses)
```
Dimosphen1: In case you have no arguments on the same line with an open parenthesis, your closing parenthesis should not have any arguments on the same line as well. Reformat your code.

Dimosphen1 (changes requested):
One minor change was requested.

app/main.py
loplicat: Thank you. I've already reformatted it.

loplicat: Also, I suppose there is a little mistake in the test in the file with the task. When it checks your code, it should include a space after "You have bought:". But tests on GitHub may not pass, because they have this string without a space

Polyakiv-Andrey (approved):
Good job!

PR #442, opened 2023-11-07
Solution

Lebwa1337:
I have some problems with path to file, so i needed 9 and 10 line

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,0 +1,10 @@
+from app.car import Car
+
+
+class Customer:
+    def __init__(self, customer: dict) -> None:
+        self.name = customer.get("name")
+        self.product_cart = customer.get("product_cart")
+        self.location = customer.get("location")
+        self.money = customer.get("money")
+        self.car = Car(customer)
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,38 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+import os.path
+
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    current_directory = os.path.dirname(os.path.abspath(__file__))
+    file_path = os.path.join(current_directory, "config.json")
+    with open(file_path, "r") as file:
+        config = json.load(file)
+        fuel_price = config.get("FUEL_PRICE")
+        for customer_ in config.get("customers"):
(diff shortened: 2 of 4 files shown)


app/main.py
```diff
@@ -1,3 +1,38 @@
...
+            customer = Customer(customer_)
+            print(f"{customer.name} has {customer.money} dollars")
+            lowest_price = 10000000
+            optimal_place = ""
+            for shop_ in config.get("shops"):
+                shop = Shop(shop_)
+                cost_of_trip = shop.calculate_trip_price(customer, fuel_price)
+                print(f"{customer.name}'s trip to the {shop.name} "
+                      f"costs {cost_of_trip}")
+                if cost_of_trip < lowest_price:
```
viktoria-rybenchuk, line 24: What if lowest cost of trip is  20000000?) You don't need hardcode `lowest_price` as  10000000. It is better create separate function for this logic. And you should operate only data that you have, don`t add additional numbers)

app/customer.py
```diff
@@ -0,0 +1,10 @@
+from app.car import Car
+
+
+class Customer:
+    def __init__(self, customer: dict) -> None:
+        self.name = customer.get("name")
+        self.product_cart = customer.get("product_cart")
+        self.location = customer.get("location")
+        self.money = customer.get("money")
+        self.car = Car(customer)
```
viktoria-rybenchuk, line 10: add `str` method in this class as well

app/main.py
```diff
@@ -1,3 +1,38 @@
...
+
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    current_directory = os.path.dirname(os.path.abspath(__file__))
+    file_path = os.path.join(current_directory, "config.json")
+    with open(file_path, "r") as file:
+        config = json.load(file)
```
viktoria-rybenchuk, line 12: Do not overload the context manager.
The context manager is needed to work with the file, in our case, to read data. Therefore, after you have read the data from the file, exit the block.

viktoria-rybenchuk (changes requested):

Lebwa1337:
Why my tests are failed?In pycharm test is passed

Dimosphen1 (approved):
LGTM, all tests passed successfully locally. Well done!

PR #444, opened 2023-11-08
Fix extra space

olena-hul:


olena-hul:
@Abnormaltype review pls

PR #449, opened 2023-11-09
solution

Stainback:


app/config.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+
+    for shop in data["shops"]:
+        SHOPS.append(
+            Shop(
+                shop["name"],
+                shop["location"],
+                shop["products"]
+            )
+        )
```
EdAlekseiev, line 36: Move this code from context manager

app/customer/car.py
```diff
@@ -0,0 +1,18 @@
...
+    fuel_price = 0
+
+    def __init__(self, brand: str, fuel_consumption: float) -> None:
+        self.brand = brand
+        self.fuel_consumption = fuel_consumption
+
+    def calculate_ride_cost_to(
+            self,
+            location: list,
+            destination: list
```
EdAlekseiev, line 14: Specify inner type of list

app/shop.py
```diff
@@ -0,0 +1,46 @@
...
+
+                receipt["total_cost"] += cost
+                receipt["cart"].append((cart[product], product, cost))
+
+        self.purchases[customer_name] = receipt
+
+        return receipt["total_cost"]
+
+    def print_receipt(self, customer_name: str) -> None:
+        date = datetime(2021, 1, 4, 12, 33, 41).strftime("%d/%m/%Y %H:%M:%S")
```
EdAlekseiev, line 34: You must display current date

app/config.py
```diff
@@ -0,0 +1,36 @@
...
+from app.shop import Shop
+
+with open("app/config.json", "r") as config_file:
+    data = json.load(config_file)
+
+    FUEL_PRICE = data["FUEL_PRICE"]
+    CUSTOMERS = []
+    SHOPS = []
+
+    Car.fuel_price = FUEL_PRICE
```
EdAlekseiev, line 13: I'm not sure that this is the best idea. You can pass fuel_price as argument to the method `calculate_ride_cost_to `

EdAlekseiev (changes requested):

app/shop.py
Stainback: Tests are made in a way that they are failed when the exact date is not given. So I will change it back...

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #458, opened 2023-11-25
Implemented py-shop-trip

Aeros72:


app/shop.py
```diff
@@ -0,0 +1,58 @@
...
+    location: list[int]
+    products: dict
+
+    def calculate_product_cost(self, customer: Customer) -> float:
+        total_price = 0
+        for product, quantity in customer.product_cart.items():
+            if product in self.products:
+                total_price += quantity * self.products[product]
+
+        return total_price
```
ArsenPidhoretskyi: Use sum function

app/shop.py
```diff
@@ -0,0 +1,58 @@
...
+    def print_purchase_receipt(self, customer: Customer) -> None:
+        actual_dt = datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+        total_price = self.calculate_product_cost(customer)
+
+        print("\nDate:", actual_dt)
+        print(f"Thanks, {customer.name}, for your purchase!\nYou have bought:")
+
+        for product, quantity in customer.product_cart.items():
+            price = quantity * self.products[product]
+            if price == int(price):
```
ArsenPidhoretskyi: use isinteger

ArsenPidhoretskyi (changes requested):
Fix tests

app/shop.py
Aeros72: I can`t do that, as i know there is know isinteger in Python, and when i try to use isinstance it gives the wrong result, because 3.0 for example is not an integer, but i need it to be an integer, I only can check if string of this number after . is 0, but it`ll make my code more complex. Maybe you know some other way how to do it?

ArsenPidhoretskyi: if price.is_integer():

https://python-reference.readthedocs.io/en/latest/docs/float/is_integer.html

ArsenPidhoretskyi (changes requested):
Explain

app/shop.py
Aeros72: Thank you!

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        customers.append(new_customer)
+
+    shops = []
+    for shop in shops_data:
+        new_shop = Shop(
+            shop["name"],
+            shop["location"],
+            shop["products"]
+        )
+        shops.append(new_shop)
```
viktoria-rybenchuk: use list comprehension

app/shop.py
```diff
@@ -0,0 +1,60 @@
...
+    def calculate_product_cost(self, customer: Customer) -> float:
+        total_price = sum(
+            quantity * self.products[product]
+            for product, quantity in customer.product_cart.items()
+        )
+
+        return total_price
+
+    def print_purchase_receipt(self, customer: Customer) -> None:
+        timestamp = 1609763621
```
viktoria-rybenchuk: why do you need this?

app/shop.py
```diff
@@ -0,0 +1,60 @@
...
+        actual_dt = (datetime.fromtimestamp(timestamp).
+                     strftime("%d/%m/%Y %H:%M:%S"))
```
viktoria-rybenchuk: you can use simple `datetime.datetime.now().strftime('%d/%m/%Y %H:%M:%S')`

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json") as file:
+        data = json.load(file)
+        fuel_price = data["FUEL_PRICE"]
+        customers_data = data["customers"]
+        shops_data = data["shops"]
+
```
viktoria-rybenchuk: Retrieve specific data from the loaded JSON outside the context manager.

viktoria-rybenchuk (changes requested):

app/main.py
viktoria-rybenchuk: not fixed

app/shop.py
viktoria-rybenchuk: not fixed

viktoria-rybenchuk (changes requested):
Make adjustments according to the comments.

app/shop.py
Aeros72: No, I can`t use it. If I use it i`ll have 26 11 and in tests there is 04 01 and my tests will be failed

viktoria-rybenchuk: Screenshot from 2023-11-26 20-50-34
Read checklist

Aeros72: I have used your variant before and tests failed, other menthor said to fix it

Aeros72: Oh, thanks, my mistake, will be more attentive in future

viktoria-rybenchuk (approved):
GJ🚀

PR #464, opened 2023-12-08
trying to solve

Yevheniia-Ilchenko:


YukihiroSM (changes requested):
Discussing in chat. https://mate.academy/chats/25647?selectedMessageId=981438

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+            fuel_consumption: float
+    ) -> float:
+        x1, y1 = self.customer_location
+        x2, y2 = shop_location
+        dist = math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
+        price_road = dist / 100 * fuel_consumption * fuel_price * 2
+        return round(price_road, 2)
+
+    def shopping_time(self, shop: Shops) -> None:
+        current_time = datetime(2021, 1, 4, 12, 33, 41)
```
YukihiroSM: You shouldn't hardcode current time. Use datetime.now for it.

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+        print(f"Thanks, {self.name}, for your purchase!")
+        print("You have bought:")
+
+        total_cost = 0
+        for product, quantity in self.products_cart.items():
+            if product in shop.products:
+                total_cost_p = shop.products[product] * quantity
+                cost_product = int(total_cost_p)\
+                    if total_cost_p == int(total_cost_p)\
+                    else total_cost_p
```
YukihiroSM: don't use `\` for line separation. it's better to use `(` `)`

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+                print(f"{quantity} {product}s for {cost_product} dollars")
+
+                total_cost += (int(cost_product)
+                               if cost_product == int(cost_product)
+                               else cost_product)
+
+        print(f"Total cost is {round(total_cost, 2)} "
+              f"dollars\n"f"See you again!")
```
YukihiroSM: W
```suggestion
              f"dollars\nSee you again!")
```

YukihiroSM (changes requested):
Also, I would reccomend you to use black for code reformatting, because there are several formatting issues.

Yevheniia-Ilchenko:
when I correcting the date hapen error. In the task: For example, let's say now is 04/01/2021 12:33:41

Yevheniia-Ilchenko:
maybe I didn’t understand what to fix

app/customer.py
```diff
@@ -0,0 +1,57 @@
+import math
+from datetime import datetime
```
YukihiroSM: Import this as just datetime
```suggestion
import datetime
```

app/customer.py
YukihiroSM: Use current_time = datetime.datetime.now()

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #467, opened 2023-12-21
Solution

lyaskov:


app/customer.py
```diff
@@ -0,0 +1,21 @@
...
+class Customer:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.product_cart = None
+        self.location = []
+        self.money = None
+        self._car = None
+
+    @property
+    def car(self) -> Car:
```
ArsenPidhoretskyi: It cam return None also

app/data_read.py
```diff
@@ -0,0 +1,38 @@
+import json
+from typing import List
+
+from app.car import Car
+from app.customer import Customer
+from app.shop import Shop
+
+
+class DataRead:
```
ArsenPidhoretskyi:
```suggestion
class DataReader:
```

ArsenPidhoretskyi: class - noun, function - verb

app/customer.py
```diff
@@ -0,0 +1,21 @@
+from app.car import Car
+
+
+class Customer:
+    def __init__(self, name: str) -> None:
```
ArsenPidhoretskyi: Pass all parameters into init. instead of initialize them with None

ArsenPidhoretskyi: The same with Shop

app/shop.py
```diff
@@ -0,0 +1,50 @@
...
+        self.name = name
+        self.location = location
+        self.products = []
+
+    def check_cost_products(self, list_products: dict) -> int | float:
+        total = 0
+        for product, count in list_products.items():
+            total += self.products[product] * count
+
+        return total
```
ArsenPidhoretskyi: Use sum function

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #471, opened 2023-12-26
Solution for py_shop_trip

GibbsVlad:


app/shop.py
```diff
@@ -0,0 +1,41 @@
...
+        self.products = shop["products"]
+
+    def customer_buying(
+            self,
+            customer: callable) -> None:
+
+        total_cost = 0
+
+        # print("Date: ", datetime.now().strftime("%d/%m/%y %H:%M:%S"))
+        print("Date: 04/01/2021 12:33:41")
```
EdAlekseiev: Use `datetime.now()` instead of hardcode

EdAlekseiev (changes requested):

app/shop.py
GibbsVlad: Understand, but how to write it correct if task force me to print exactly 04/01/2021 12:33:41?

vasylhnatiuk (approved):

PR #485, opened 2024-01-24
Solution

kabust:


ArturPoltser (approved):

app/data/json_extraction.py
```diff
@@ -0,0 +1,12 @@
...
+
+def extract_data() -> dict:
+    with open("app/data/config.json", "r") as data_in:
+        return json.load(data_in)
+
+
+data = extract_data()
+FUEL_PRICE = data["FUEL_PRICE"]
+customers = data["customers"]
+shops = data["shops"]
```
NEkropo1: you can import this function in main and  create constants there, instead of importing them, that's more explicit way

app/objects/customer.py
```diff
@@ -0,0 +1,46 @@
...
+        self.car = car
+
+    def trip_cost(self, shop: Shop) -> float:
+        distance = ((self.location[0] - shop.location[0]) ** 2
+                    + (self.location[1] - shop.location[1]) ** 2) ** .5
+        fuel_cost = ((self.car.fuel_consumption / 100)
+                     * FUEL_PRICE * distance * 2)
+        cart_total = sum(self.product_cart[product] * shop.products[product]
+                         for product in self.product_cart.keys())
+        return round(fuel_cost + cart_total, 2)
```
NEkropo1: better to make explicit function for each calculation.

app/objects/customer.py
```diff
@@ -0,0 +1,46 @@
...
+            if product_total == int(product_total):
+                product_total = int(product_total)
+            print(f"{amount} {product}s for "
+                  f"{product_total} dollars")
+
+        cart_total = sum(self.product_cart[product] * shop.products[product]
+                         for product in self.product_cart.keys())
+
+        print(f"Total cost is {cart_total} dollars")
+        print("See you again!\n")
```
NEkropo1: This seems more like shop business logic.

app/customers_list.py
```diff
@@ -0,0 +1,18 @@
...
+        customer["product_cart"],
+        customer["location"],
+        customer["money"],
+        Car(
+            customer["car"]["brand"],
+            customer["car"]["fuel_consumption"]
+        )
+    )
+    for customer in customers
+]
```
NEkropo1: pass object creation to main, we anyway have customer data, so we kinda dont need customers_list as a python module, when all responsibility is creating one list with customers.

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+    for customer in customers:
+        print(f"{customer.name} has {customer.money} dollars")
+
+        min_trip_cost = float("inf")
+        shop_choice = None
+
+        for shop in shops:
+            trip_cost = customer.trip_cost(shop)
+            print(f"{customer.name}'s trip to the "
+                  f"{shop.name} costs {trip_cost}")
```
NEkropo1: I guess it can be a part of customers logic when he tries to buy products.

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+                  f"money to make a purchase in any shop")
+            return
+
+        print(f"{customer.name} rides to {shop_choice.name}")
+        customer.print_receipt(shop_choice)
+
+        print(f"{customer.name} rides home")
+
+        customer.money -= min_trip_cost
+        print(f"{customer.name} now has {customer.money} dollars\n")
```
NEkropo1: these all also can be passed to business logic of our classes, there's no need to make it in main.

NEkropo1 (changes requested):
Try to remake app using logic from real world.

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+]
+
+shops = [
+    Shop(
+        shop["name"],
+        shop["location"],
+        shop["products"]
+    )
+    for shop in shops
+]
```
Serhii-Leonenko: data extracting and instances creating logic better to put inside the function, it's better when this logic will be executed when we call the function instead of importing

app/objects/customer.py
```diff
@@ -0,0 +1,44 @@
...
+        cart_total = self.calculate_cart_total(shop)
+        return round(fuel_cost + cart_total, 2)
+
+    def trip_choice(self, shop: Shop, fuel_price: float) -> None:
+        trip_cost = self.trip_cost(shop, fuel_price)
+        print(f"{self.name}'s trip to the "
+              f"{shop.name} costs {trip_cost}")
+        if trip_cost < self.min_trip_cost:
+            self.min_trip_cost = trip_cost
+            self.shop_choice = shop
```
Serhii-Leonenko: it's not a very good idea to store this info directly in the object in this case, because when we try to get these attributes and forgot to call this method before we cannot understand there is no cheapest trip or we just forgot to call the method. Will be better if you return these data

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+
+def shop_trip() -> None:
+    for customer in customers:
+        print(f"{customer.name} has {customer.money} dollars")
+
+        for shop in shops:
+            customer.trip_choice(shop, FUEL_PRICE)
```
Serhii-Leonenko: let this method take a list of shops, and return the cheapest shop and trip cost

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+        if customer.money < customer.min_trip_cost:
+            print(f"{customer.name} doesn't have enough "
+                  f"money to make a purchase in any shop")
+            return
```
Serhii-Leonenko: and then check the result instead of customer attributes

Serhii-Leonenko (changes requested):

Y-Havryliv (approved):

PR #491, opened 2024-01-29
Solution

u123dev:


app/customer.py
```diff
@@ -0,0 +1,40 @@
+from dataclasses import dataclass, field
+from app.car import Car, FuelPrice
+from app.shop import Shop
+
+
+@dataclass
```
EdAlekseiev, line 6: This class doesn't look as dataclass, because it contains logic too

app/shop.py
```diff
@@ -0,0 +1,25 @@
...
+@dataclass
+class Shop():
+    name: str
+    location: list[int, int]
+    products: dict
+
+    def purhase(self, customer_name: str, products: dict) -> float:
+        #  current_datetime = datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+        #  print(f"Date: {current_datetime}")
+        print("Date: 04/01/2021 12:33:41")
```
EdAlekseiev: Don't hardcode Date value. Use datetime.now()

EdAlekseiev (changes requested):

app/shop.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

YukihiroSM (approved):
Looks good to me, but a small comment for future: 
You should use dataclasses only for classes, which contain and represent data. It's better not to use usual class with complex logic as dataclass.

PR #505, opened 2024-02-12
Shop trip

Ananiev-Vitalii:


app/customer.py
```diff
@@ -0,0 +1,38 @@
...
+from app.utils import calculate_distance
+
+
+class Customer:
+    def __init__(
+            self, name: str,
+            product_cart: dict,
+            location: list,
+            money: float,
+            car: "Car") -> None:
```
vasylhnatiuk, line 12:
```suggestion
            car: Car) -> None:
```

app/customer.py
```diff
@@ -0,0 +1,38 @@
...
+    def calculate_trip_cost(
+            self, distance: float,
+            fuel_price: float) -> float:
+        fuel_consumption_per_km = self.car.fuel_consumption / 100
+        return distance * fuel_consumption_per_km * fuel_price * 2
+
+    def has_enough_money(self, cost: float) -> bool:
+        return self.money >= cost
+
+    def make_purchase(self, shop: "Shop", fuel_price: float) -> tuple:
```
vasylhnatiuk, line 28:
```suggestion
    def make_purchase(self, shop: Shop, fuel_price: float) -> tuple:
```

app/utils.py
```diff
@@ -0,0 +1,4 @@
+def calculate_distance(location1: list, location2: list) -> float:
```
vasylhnatiuk, line 1:
```suggestion
def calculate_distance(location1: list[int], location2: list[int]) -> float:
```

vasylhnatiuk: do it everywhere

app/customer.py
vasylhnatiuk: do it everywhere

vasylhnatiuk (changes requested):

PR #510, opened 2024-02-13
'Solution'

sofiazholud:


app/customer.py
```diff
@@ -0,0 +1,74 @@
...
+                f'{datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")}'
+            )
+            print(f"Thanks, {self.name}, for your purchase!")
+            print("You have bought:")
+            for product, quantity in self.product_cart.items():
+                product_price = (
+                    cheapest_shop.products[product] * quantity
+                )
+                product_price = int(product_price) \
+                    if product_price.is_integer() else f"{product_price:.2f}"
```
EdAlekseiev: product_price can be int here. Change you check

app/customer.py
```diff
@@ -0,0 +1,74 @@
...
+                print(f"{quantity} {product}s for {product_price} dollars")
+            cheapest_cost = int(cheapest_cost) \
+                if cheapest_cost.is_integer() else f"{cheapest_cost:.2f}"
```
EdAlekseiev: The same comment as above

EdAlekseiev (changes requested):

vasylhnatiuk (changes requested):
fix flake8 tests

YukihiroSM (approved):
Looks nice to me!

PR #515, opened 2024-02-17
'Solution'

JohnnyWalker010:


achernyshovv (approved):

NEkropo1 (changes requested):
I noticed your main function handles a lot of functionalities that could be more effectively managed by the classes you've defined. Here are some specific suggestions to refactor your code:

1. Methods in Classes: For instance, the code block where you print a receipt can be turned into a `print_receipt` method in the Shop class. Similarly, consider adding a `ride_with_car` method in the Customer class. This approach helps in organizing code logically and makes it reusable.

2. Use of Data Classes: If a class is only holding data and you don't plan to add methods, Python's dataclasses might be a suitable choice. They provide a neat and efficient way to create data-only classes.

3. DRY Principle: Try to identify any repeated code blocks and move them into separate methods or functions. DRY (Don't Repeat Yourself) is a principle aimed at reducing repetition in code, which makes your code easier to maintain and modify.

JohnnyWalker010:
Thanks! That makes a lot of sense and improves my code) The only thing I didn't succeed in was bringing the ride_with_car method into the Customer class because of the Circular Import problem. Print_receipt uses Shop and Customer objects same as ride_with_car. I've created the ride_with_car function, but put it into a separate module and it looks like I followed the DRY principle. Besides all classes that have to be data classes are converted into data classes.

app/customer.py
```diff
@@ -0,0 +1,28 @@
...
+    customers[customer["name"]] = Customer(
+        customer["name"],
+        customer["product_cart"],
+        customer["location"],
+        customer["money"],
+        Car(
+            customer["car"]["brand"],
+            customer["car"]["fuel_consumption"]
+        )
+    )
```
Serhii-Leonenko, line 28: it's not the best choice to put the code directly in module, because this logic will be executed every time you import the module

app/trip.py
```diff
@@ -0,0 +1,56 @@
+import math
+from app.customer import Customer
+from app.shop import Shop, shops
+from app.get_info import info_from_file
+
+
+def calculate_cost_of_trip(
+        customer: Customer,
+        shop: Shop
+) -> None:
```
Serhii-Leonenko, line 10: does it return None?

app/trip.py
```diff
@@ -0,0 +1,56 @@
+import math
+from app.customer import Customer
+from app.shop import Shop, shops
+from app.get_info import info_from_file
```
Serhii-Leonenko, line 4: check your imports that comply with the PEP8 rules

app/trip.py
```diff
@@ -0,0 +1,56 @@
...
+        2
+        * distance
+        * customer.car.fuel_consumption
+        / 100
+        * info_from_file["FUEL_PRICE"], 2
+    )
+    cart_cost = 0
+    for product_name, product_cost in customer.product_cart.items():
+        cart_cost += product_cost * shop.products[product_name]
+    return fuel_cost + cart_cost
```
Serhii-Leonenko, line 25: I recommend you to separate for loops, if conditions, return, etc with a blank line, it will increase readability of your code

app/customer.py
Serhii-Leonenko: the same situation you have in other modules, put parsing logic in the func or create parser class

Serhii-Leonenko (changes requested):

app/trip.py
JohnnyWalker010: I agree, that makes things better

app/trip.py
JohnnyWalker010: Done

Now standard library imports are separated from local imports

PR #517, opened 2024-02-18
Solution

mykytaso:


ArsenPidhoretskyi (changes requested):
Why do you use suffix Class in class names???

ArsenPidhoretskyi (approved):
Well done!

PR #523, opened 2024-02-20
Created an application that helps customers to choose the cheapest tr…

AndriiHomeniuk1:
…ip for the products.

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+from app.car import Car
+from app.trip import Trip
+from typing import List
+
+
+class Customer:
+
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: List[int],
+            money: int,
+            car: Car
+    ) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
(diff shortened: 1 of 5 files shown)


app/main.py
```diff
@@ -1,3 +1,88 @@
...
+import json
+import math
+
+from app.customer import Customer
+from app.car import Car
+from app.shop import Shop
+from app.trip import Trip
+
+
+def shop_trip() -> None:
```
vasylhnatiuk, line 10: try to separate this function to a few functions. because it is really long

app/customer.py
```diff
@@ -0,0 +1,57 @@
...
+                f"{shop.shop_name} costs {round(shop.total_trip, 2)}"
+            )
+
+        if customer.money < cheapest_trip.total_trip:
+            print(
+                f"{customer.name} doesn't have enough "
+                f"money to make a purchase in any shop")
+            return False
+
+        print(f"{customer.name} rides to {cheapest_trip.shop_name}\n")
```
vasylhnatiuk, line 43: when customer rides shop his location should be changed. same with home

vasylhnatiuk (changes requested):

PR #525, opened 2024-02-24
Created an application that helps customers to choose the cheapest tr…

AndriiHomeniuk1:
…ip for the products.

### DIFF
--- a/app/open_file_config.py
+++ b/app/open_file_config.py
@@ -1,5 +1,5 @@
 import json
-from typing import Tuple, List
+from typing import Tuple, List, Dict
 from app.car import Car
 from app.customer import Customer
 from app.shop import Shop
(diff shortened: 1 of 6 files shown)


AndriiHomeniuk1:
I completed the task taking into account the mentor’s feedback.
ментор

app/shop.py
```diff
@@ -0,0 +1,57 @@
...
+        self.products = products
+
+    @staticmethod
+    def convert_to_int_if_possible(number: float) -> float | int:
+        if number == int(number):
+            return int(number)
+        return number
+
+    @staticmethod
+    def shop_receipt(
```
vsmutok:
```suggestion
    def shop_receipt(
```

app/shop.py
```diff
@@ -0,0 +1,57 @@
...
+        customer: Customer,
+        trip_instances: List[Trip]
+    ) -> None:
+        cheapest_trip = Trip.cheapest_trip(trip_instances)
+        datetime_mock = MagicMock(wrap=datetime)
+        datetime_mock.now.return_value = datetime(2021, 1, 4, 12, 33, 41)
+
+        milk_cost = Shop.convert_to_int_if_possible(
```
vsmutok:
```suggestion
        milk_cost = self.convert_to_int_if_possible(
```

app/trip.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @staticmethod
+    def trip_distance(
+        shop_location: List[int],
+        customer_location: List[int]
+    ) -> float:
+        return math.sqrt(
+            (shop_location[0] - customer_location[0]) ** 2
+            + (shop_location[1] - customer_location[1]) ** 2
+        )
```
vsmutok: you can use `math.dist`

app/trip.py
```diff
@@ -0,0 +1,112 @@
...
+
+                total_milk_cost = (
+                    customer.product_cart["milk"] * shop.products["milk"]
+                )
+                total_bread_cost = (
+                    customer.product_cart["bread"] * shop.products["bread"]
+                )
+                total_butter_cost = (
+                    customer.product_cart["butter"] * shop.products["butter"]
+                )
```
vsmutok: hardcode 

what if customer need another product?

app/shop.py
vsmutok: it is better to operate at the level of class instances than at the level of the classes

app/shop.py
```diff
@@ -0,0 +1,57 @@
...
+            cheapest_trip.total_milk_cost
+        )
+        bread_cost = Shop.convert_to_int_if_possible(
+            cheapest_trip.total_bread_cost
+        )
+        butter_cost = Shop.convert_to_int_if_possible(
+            cheapest_trip.total_butter_cost
+        )
```
vsmutok: hardcode 

What if some customer need fish or chicken for example?

app/shop.py
vsmutok: fix it wherever you can

app/trip.py
```diff
@@ -0,0 +1,112 @@
...
+            * trip_distance / 100
+        ) * 2
+
+    @staticmethod
+    def choose_shop_and_go_to_trip(
+        customer_instances: list,
+        shop_instances: list
+    ) -> None:
+        class_customer = import_module("app.customer").Customer
+        class_shop = import_module("app.shop").Shop
```
vsmutok: Why you need it?

app/open_file_config.py
```diff
@@ -0,0 +1,41 @@
...
+                customer_instances.append(customer_instance)
+
+            for shop in data_config.get("shops"):
+                shop_instance = Shop(
+                    name=shop["name"],
+                    location=shop["location"],
+                    products=shop["products"]
+                )
+                shop_instances.append(shop_instance)
+        return customer_instances, shop_instances
```
vsmutok: better to create one method for `open_file_config` and other method for instances creations

vsmutok (changes requested):

app/open_file_config.py
AndriiHomeniuk1: yes, I did it

app/trip.py
AndriiHomeniuk1: Yes, I understand)) 
I fixed it

app/shop.py
AndriiHomeniuk1: fixed it

app/shop.py
AndriiHomeniuk1: yes, that's much better) 
I fixed it where I could

app/trip.py
AndriiHomeniuk1: much more convenient
thank you)

app/trip.py
AndriiHomeniuk1: "I used importlib because there was an error with cyclic imports of the Customer and Shop modules.
But now I fixed it too

ArsenPidhoretskyi (approved):
Good job

PR #551, opened 2024-04-08
Solution

vitalinamal:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+def shop_trip() -> None:
+
+    with open("app/config.json", "r") as file:
+        config = json.load(file)
+    fuel_price = config["FUEL_PRICE"]
+
+    shops = [Shop(**shop) for shop in config["shops"]]
+    customers = [Customer(**{**customer, "car": Car(**customer["car"])})
+                 for customer in config["customers"]]
```
LiudmylaKulinchenko: U can perform car creation inside `Customer.__init__` method

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+            trip_cost = customer.calculate_trip_cost(shop.location, fuel_price)
+            products_cost = shop.calculate_products_cost(customer.product_cart)
+            total_cost = trip_cost + products_cost
+
+            print(f"{customer.name}'s trip to the {shop.name} "
+                  f"costs {total_cost:.2f}")
+
+            if total_cost < customer.money and total_cost < minimum_total_cost:
+                cheapest_shop = shop
+                minimum_total_cost = total_cost
```
LiudmylaKulinchenko: You can create a separate function for this

LiudmylaKulinchenko (changes requested):
try to separate shop_trip logic into smaller functions, SRP

app/main.py
vitalinamal: Inside README.md says: "The main.py module must contain only shop_trip function". Where I should put this function?

.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 inline-quotes = "
-ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
+ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818, E231
```
Oleksl888: you can add # noqa: E231 against the line to silence that warning

tests/config.json
```diff
@@ -0,0 +1,76 @@
+{
+    "FUEL_PRICE": 2.4,
```
Oleksl888: why did you move this file here?

app/main.py
Oleksl888: You can put all logic that makes the shop trip in a customer class and then all you do in shop_trip() function is to call this method on customer instance

Oleksl888: A good design here would be to keep only reading json, creating customers and calling customer methods here. This way you will structure your code better.

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        config = json.load(file)
+
+    fuel_price = config["FUEL_PRICE"]
+    shops = [Shop(**shop) for shop in config["shops"]]
+    customers = [Customer(**customer) for customer in config["customers"]]
```
Oleksl888: you can create a customer in a loop on line 15 and call .go_shopping() there - thus you will not need to do two iterations

Oleksl888 (changes requested):

app/customer.py
```diff
@@ -0,0 +1,64 @@
...
+        cheapest_shop = None
+        minimum_total_cost = float("inf")
+
+        for shop in shops:
+            trip_cost = self.calculate_trip_cost(shop.location, fuel_price)
+            products_cost = shop.calculate_products_cost(self.product_cart)
+            total_cost = trip_cost + products_cost
+
+            print(f"{self.name}'s trip to the {shop.name} "
+                  f"costs {total_cost:.2f}")  # noqa: E231
```
Oleksl888, line 42: if the tests results conflict with flake8 please report this on the platform

Oleksl888 (approved):
very nice!

PR #554, opened 2024-04-10
Solution

kolya2731:


app/shop.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def calculate_total_price(self) -> float | int:
+        return sum(
+            self.shop_products[product] * value
+            for product, value in self.products_cart.items()
+        )
+
+    def print_account(self) -> None:
+        print(f"Date: "
+              f"{datetime(2021, 4, 1, 12, 33, 41).strftime('%m/%d/%Y %T')}")
```
pavlopro: datetime.datetime.now()

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+                print(f"{customer['name']}'s trip to the "
+                      f"{shop['name']} costs {car + sh}")
+
+            if customer["money"] < min(res.keys()):
+                print(f"{customer['name']} doesn't have enough "
+                      f"money to make a purchase in any shop")
+                break
+
+            print(f"{customer['name']} rides to {res[min(res)][-1]}\n")
```
pavlopro: When the customer arrives at the shop his location should equal to shop location. The same when he arrives home.

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+            for shop in config["shops"]:
+
+                car = Car(
+                    config["FUEL_PRICE"],
+                    customer["car"]["fuel_consumption"],
+                    customer["location"], shop["location"]
+                ).calculate_trip_cost()
+
+                sh = Shop(
```
pavlopro: Please use more descriptive variable names. The current names 'car' and 'sh' do not adequately convey that they are related to prices. Consider renaming them to reflect their association with pricing

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+                    shop["products"], shop["name"]
+                ]
+
+                print(f"{customer['name']}'s trip to the "
+                      f"{shop['name']} costs {car + sh}")
+
+            if customer["money"] < min(res.keys()):
+                print(f"{customer['name']} doesn't have enough "
+                      f"money to make a purchase in any shop")
+                break
```
pavlopro: Continue instead of break, because function will stop checking the remaining customers

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+            print(f"{customer['name']} rides to {res[min(res)][-1]}\n")
+
+            Shop(
+                res[min(res)][0],
```
pavlopro: Consider creating more readable variable names, as it's difficult to understand what res[min(res)][0] and similar expressions represent.

app/shop.py
```diff
@@ -0,0 +1,35 @@
...
+        print(f"Thanks, {self.name_customer}, for your purchase!")
+        print("You have bought:")
+        for product, value in self.products_cart.items():
+            product_s = self.shop_products[product]
+            if product_s * value == int(product_s * value):
```
pavlopro: Consider simplifying the code to avoid duplicate printing of similar data.

pavlopro (changes requested):

app/shop.py
kolya2731: when i use datetime.now() tests doesn't pass

app/shop.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def calculate_total_price(self) -> float | int:
+        return sum(
+            self.shop_products[product] * value
+            for product, value in self.products_cart.items()
+        )
+
+    def print_account(self) -> None:
+        print(f"Date: "
+              f"{datetime(2021, 4, 1, 12, 33,41).strftime('%m/%d/%Y %T')}")
```
pavlopro, line 19: Try this

print(f"Date: {datetime.now().strftime('%d/%m/%Y %H:%M:%S')}")

pavlopro (changes requested):

app/shop.py
kolya2731: Знімок екрана 2024-04-10 163222

Polyakiv-Andrey (approved):
Good job!

PR #555, opened 2024-04-10
solution

alonasorochynska:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+import json
+import math
+
+
+from app.customer import Customer
+from app.shop import Shop, print_check
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as f:
```
Oleksl888: 1.Think how you can structure your code better. A good solution would be to have about 10 lines of code in this function. Here you should take json file, create customers and call necessary methods on customer to make a shop trip.
2.You should create more classes (e.g. Car, Location) they will all have some methods that you can reuse in other classes to complete this task (for example calculate car trip cost and calculate distance between two locations).
3. Add methods to your customer class - you should write methods here that will simulate a shop trip.
4. Also your tests are failing so keep working on it :)

Oleksl888 (changes requested):

alonasorochynska:
I can not pass 1 one pytest because of an issue on screenshot. And I dont know how to solve this 
image

Oleksl888:
> I can not pass 1 one pytest because of an issue on screenshot. And I dont know how to solve this
> image

Maybe try to call int() on a cost value but only when it ends up it’s .0
This is the костиль but what to do

app/customer.py
```diff
@@ -0,0 +1,76 @@
...
+@dataclass
+class Customer:
+    name: str
+    product_cart: dict
+    location: list
+    money: int
+    car: dict
+
+
+def cost_of_km(customer: Customer, data: dict) -> int | float:
```
Oleksl888: This method logically belongs to class Car. Also you can pass here data["FUEL_PRICE"] - as a single element, no need to pass entire dict

app/customer.py
```diff
@@ -0,0 +1,76 @@
...
+    return (customer.car["fuel_consumption"] / 100) * data["FUEL_PRICE"]
+
+
+def cost_by_distance(
```
Oleksl888: you should rewrite these functions into methods and then call them on the instance of a class

app/customer.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def total_price(customer: Customer, shop: Shop, data: dict) -> int | float:
+    return cost_by_distance(
+        customer, shop, data) + cost_of_products(
+        customer, shop)
+
+
+def trip_and_purchase(data: dict) -> None:
+    for customer_data in data["customers"]:
```
Oleksl888: The part of logic that creates instances of Customer and Shop can be moved to main.py. There you can create instances and then call methods to perform a shop trip on a customer instance

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as config_file:
+        data = json.load(config_file)
+        customers = data["customers"]
+        shops = data["shops"]
+
+    for person in customers:
```
Oleksl888, line 14: So again this main() function is way too big. A better implementation would be like:
1. Make a car as one of the attrs of the customer (would make sense because they are kind of linked in real world)
2. Create shop instances in list comprehension.
3. Move everything below line 32 into a separate method or methods on customer (e.g. find_cheapest_store_and_go_there).
4. Call this method on each newly created instance of a customer.
This way you will make your main function cleaner - and break down logic of code better

Oleksl888 (changes requested):

PR #556, opened 2024-04-10
Solution

dkibalenko:


app/main.py
```diff
@@ -1,3 +1,44 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    json_file_path = ("C:/Users/Dennie/mate-projects/"
```
ArsenPidhoretskyi: Add relative path
When i try to run this in my local computer it does not work....

app/customer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def distance_to(self, shop: Shop) -> float:
+        return math.sqrt(
+            (self.location[0] - shop.location[0])
+            ** 2 + (self.location[1] - shop.location[1]) ** 2
+        )
+
+    def count_fuel_cost(self, shop: Shop, fuel_price: float) -> float:
+        if not isinstance(shop, Shop):
+            raise TypeError(f"{shop} should be Shop type")
```
ArsenPidhoretskyi: Due to type annotations it is always true
So two options:
1) Change type annotations
2) Remove redundant if condition

Please, fix in all places

ArsenPidhoretskyi (changes requested):

app/customer.py
```diff
@@ -0,0 +1,76 @@
+import math
+import datetime
```
Dimosphen1, line 2: Add a blank line between dfferent import groups to follow PPEP-8. Please, apply this change to all places

Dimosphen1 (changes requested):
Several minor changes were requested

vasylhnatiuk (approved):

PR #558, opened 2024-04-15
Solution for task

oleksandra-shershen:


Oleksl888 (approved):
Overall good job! 
I would only suggest to make your code more modular, for example:
1. Create class Location and move method `calculate_distance` there.
2. `calculate_fuel_consumption` could be a part of a Car class - and then you could call it on customer with customer.car...
This does not affect functionality here, but will make your code better structured. It is a good practice to avoid having one class that does-it-all with lots of methods but to distribute logic between different logical units

app/shop.py
```diff
@@ -0,0 +1,25 @@
...
+        return sum(self.products[product] * quantity
+                   for product, quantity in product_cart.items())
+
+    def print_receipt(
+            self,
+            customer_name: str,
+            product_cart: dict,
+            products_cost: float
+    ) -> None:
+        print(f"\nDate: 04/01/2021 12:33:41\n"
```
Oleksl888, line 17: This should not be hard coded

PR #563, opened 2024-04-18
Solution

maxbidun:
I have problem with datetime() can you help me?

### DIFF
--- a/app/shop.py
+++ b/app/shop.py
@@ -14,12 +14,9 @@ class Shop:
         self.time = datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")
 
     def price_products(self, customer_product: Customer) -> int:
-        for product in customer_product.product_cart:
-            price_for_product = (customer_product.product_cart[product]
-                                 * self.products[product])
-            self.dict_products.update({product: [
-                price_for_product,
-                customer_product.product_cart[product]]})
+        for product, price in customer_product.product_cart.items():
+            price_for_product = (price * self.products[product])
+            self.dict_products.update({product: [price_for_product, price]})
             self.cost += price_for_product
         distance = Distance(customer_product, self.location)
         self.cost += distance.distance_between()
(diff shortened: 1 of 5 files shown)


app/shop.py
```diff
@@ -0,0 +1,28 @@
+from app.customer import Customer
+from app.distance import Distance
+from datetime import datetime
```
ArsenPidhoretskyi: Use:
import datetime
And than into code: datetime.datetime.now().... and it will work as expected(for tests)

app/shop.py
```diff
@@ -0,0 +1,28 @@
...
+        self.name = shop["name"]
+        self.location = shop["location"]
+        self.products = shop["products"]
+        self.dict_products = {}
+        self.cost = 0
+        self.distance = 0
+        self.time = datetime.now().strftime("%m/%d/%Y %H:%M:%S")
+
+    def price_products(self, customer_product: Customer) -> int:
+        for i in customer_product.product_cart:
```
ArsenPidhoretskyi: i - is bad variable name, you can also iterate over .items()

ArsenPidhoretskyi (changes requested):

app/shop.py
maxbidun: thank you

app/shop.py
```diff
@@ -0,0 +1,29 @@
...
+        self.name = shop["name"]
+        self.location = shop["location"]
+        self.products = shop["products"]
+        self.dict_products = {}
+        self.cost = 0
+        self.distance = 0
+        self.time = datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+
+    def price_products(self, customer_product: Customer) -> int:
+        for product in customer_product.product_cart:
```
ArsenPidhoretskyi: Iterate over .items() instead of keys
for product, price in customer_product.product_cart.items()

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #567, opened 2024-04-25
solution

ArtemLeo:


app/main.py
```diff
@@ -1,3 +1,35 @@
-def shop_trip():
-    # write your code here
-    pass
+import json
+import os.path
+
```
Dimosphen1, line 3: There is no need to place a blank line before the `decimal` import since it belongs to the same group

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #582, opened 2024-06-03
"Sol P"

poplyasita:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+from app.car import Car
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open(
+            "/Users/olenapoplias/PycharmProjects/py-shop-trip/app/config.json",
+            "r"
```
viktoria-rybenchuk: When running tests on GitHub, there is no need to use a local absolute path, as it cannot find your directory. Instead, use `app/config.json`

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #586, opened 2024-06-06
py-shop-trip v1

BornToLivee:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+            print(f"{customer.name} doesn't have enough money"
+                  f" to make a purchase in any shop")
+            continue
+
+        milk_price = (customer.product_cart["milk"]
+                      * cheapest_shop.products["milk"])
+        bread_price = (customer.product_cart["bread"]
+                       * cheapest_shop.products["bread"])
+        butter_price = (customer.product_cart["butter"]
+                        * cheapest_shop.products["butter"])
```
DevSpaciX: DRY ! Create func for reuse your code

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    all_data = json.load(data_file)
+
+customers = []
+for customer_data in all_data["customers"]:
+    name = customer_data["name"]
+    product_cart = customer_data["product_cart"]
+    location = customer_data["location"]
+    money = customer_data["money"]
+    car = customer_data["car"]
+    customer = Customer(name, product_cart, location, money, car)
```
DevSpaciX: You can use @dataclass for this logic . Read more about it ) This is really useful

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    customers.append(customer)
+shops = []
+for shop_data in all_data["shops"]:
+    name = shop_data["name"]
+    location = shop_data["location"]
+    products = shop_data["products"]
+    shop = Shop(name, location, products)
+    shops.append(shop)
```
DevSpaciX: Same here ) @dataclass must have for this operations

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+def calculate_product_cost(product_cart: dict, products: dict) -> float:
+    return sum(product_cart[product] * products[product]
+               for product in product_cart)
+
+
+def format_price(price: float) -> str:
+    if price % 1 == 0:
+        return str(int(price))
+    else:
+        return f"{price}"
```
DevSpaciX: I think you can remove this method and use `round(..., 2)` only . 
Also second return is redundat here

DevSpaciX (changes requested):

DevSpaciX (approved):
Next time separate 1 big func into a smaller by logic

PR #590, opened 2024-06-18
Solution

lilarin:


app/trip.py
```diff
@@ -0,0 +1,64 @@
...
+        distance_price = (distance / 100) * fuel_consumption * fuel_price * 2
+        product_price = 0
+
+        for product, amount in customer.product_cart.items():
+            cost = shop.products[product] * amount
+            product_price += cost
+
+        if distance_price + product_price < min_price:
+            min_price = distance_price + product_price
+            min_shop = shop
```
DevSpaciX: Separate all calculation logic into different functions . That will be more clean imho

app/trip.py
```diff
@@ -0,0 +1,64 @@
...
+    if min_price > customer.money:
+        print(f"{customer.name} doesn't have enough money "
+              f"to make a purchase in any shop")
+        return
+
+    print(f"{customer.name} rides to {min_shop.name}\n")
+
+    current_date = datetime(
+        2021, 1, 4, 12, 33, 41
+    ).strftime("%d/%m/%Y %H:%M:%S")
```
DevSpaciX: Can be this a global viriable ?

DevSpaciX (changes requested):
Break your trip() function into several logical ones. And notice that func() that bigger than 30 rows , it is not by pep8 conditions

DevSpaciX (approved):
Good job!

PR #598, opened 2024-07-18
Solution

Wallian169:


app/main.py
```diff
@@ -1,3 +1,52 @@
-def shop_trip():
-    # write your code here
-    pass
+from app.customer import Customer
+from app.shop import Shop
+import json
```
Dimosphen1: Order and group your imports to follow PEP-8 import rules here and in the other files

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
```
Dimosphen1: Use `os.path.join` to join directories and make the code compatible with different operating systems

Dimosphen1 (changes requested):
Several changes were requested

pavlopro (approved):
Good Job!

PR #600, opened 2024-07-19
implemented "Shop trip" behaviour

MasakDirt:


app/main.py
```diff
@@ -1,3 +1,91 @@
...
+
+    return customers
+
+
+def get_instance_of_car(car: dict) -> Car:
+    return Car(brand=car["brand"], fuel_consumption=car["fuel_consumption"])
+
+
+def init_shops(data: list) -> list[Shop]:
+    result_list = []
```
Oleksl888: you can use list comprehension here

app/main.py
```diff
@@ -1,3 +1,91 @@
...
+            min_cost = value
+            current_shop = key
+
+    return [min_cost, current_shop]
+
+
+def shop_trip() -> None:
+    data = get_data_from_json()
+    fuel_price = data["FUEL_PRICE"]
+    customers = init_customers(data["customers"])
```
Oleksl888: you can optimize your code so you don't have to iterate over the same data twice. You can basically create customer and do all the trip related actions in one loop

app/customer.py
```diff
@@ -0,0 +1,62 @@
...
+        for shop in shops:
+            trip_cost = self.calculate_trip_cost(shop, fuel_price)
+            print(f"{self.name}'s trip to the {shop.name} costs {trip_cost}")
+            result_costs[shop] = trip_cost
+
+        return result_costs
+
+    def calculate_trip_cost(self, shop: Shop, fuel_price: float) -> float:
+        distance_to_shop_and_home = sqrt(
+            (shop.location[0] - self.location[0]) ** 2
```
Oleksl888: You could create class Location with method that calculates distance between two points to better structure your code

app/customer.py
```diff
@@ -0,0 +1,62 @@
...
+            + (shop.location[1] - self.location[1]) ** 2)
+
+        fuel_cost_to_shop_and_home = (distance_to_shop_and_home
```
Oleksl888: this could be a Car method

app/customer.py
```diff
@@ -0,0 +1,62 @@
...
+                                      * (self.car.fuel_consumption / 100)
+                                      * fuel_price)
+
+        product_cost_total = sum(shop.products[key] * value
```
Oleksl888: This could be a method defined in Shop

app/main.py
```diff
@@ -1,3 +1,91 @@
...
+
+        min_cost = min_cost_and_current_shop_list[0]
+        current_shop = min_cost_and_current_shop_list[1]
+
+        if customer.money < min_cost:
+            print(f"{customer.name} doesn't have enough"
+                  f" money to make a purchase in any shop")
+            return
+
+        print(f"{customer.name} rides to {current_shop.name}")
```
Oleksl888: you can put all this prints as a method in your Customer class - this way you will offload a bunch of code from your main function and make it cleaner

app/customer.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __repr__(self) -> str:
+        return (f"Customer(name={self.name}, products={self.product_cart}, "
+                f"location={self.location}, "
+                f"money={self.money}, car={self.car})")
+
+    def buy_products(self, current_shop: Shop) -> float:
+        total_cost = 0
+        for key, value in self.product_cart.items():
+            for _ in range(3):
```
Oleksl888: why do you need two for loops here?

app/customer.py
```diff
@@ -0,0 +1,62 @@
...
+                cost = current_shop.products[key] * value
+                if ".0" in str(cost):
```
Oleksl888: save `cost` using ternary operator and then pass it to print only once like so:
```
product_cost = ... if something else ...
print(f"{value} {key}s for {product_cost} dollars")

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌸

PR #605, opened 2024-07-27
Implemented solution

dimak20:


app/customer.py
```diff
@@ -0,0 +1,24 @@
+class Customer:
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: list,
+            money: int | float,
```
oliasymak24, line 14: For money usually better to use Decimal module

app/customer.py
```diff
@@ -0,0 +1,24 @@
...
+            car: dict
+    ) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.home_location = location
+        self.money = money
+        from app.car import Car
```
oliasymak24: Put on the beginning

app/shop.py
```diff
@@ -0,0 +1,39 @@
...
+from app.customer import Customer
+
+
+class Shop:
+    def __init__(self, name: str, location: list, products: dict) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def calculate_purchase(self, customer: Customer) -> int | float:
```
oliasymak24: Fix annotation with Decimal

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            print(
+                f"{customer.name}'s trip to the "
+                f"{shop.name} costs {all_cost}"
+            )
+            if min_cost == 0:
+                min_cost = all_cost
+                shop_index = index
+            if all_cost < min_cost:
+                min_cost = all_cost
+                shop_index = index
```
oliasymak24: I think this will be more optimized if min_cost == 0 or all_cost < min_cost

oliasymak24 (changes requested):

app/customer.py
```diff
@@ -0,0 +1,55 @@
...
+    def arrive_at_shop(self, shop: object) -> str:
+        self.location = shop.location
+        return f"{self.name} rides to {shop.name}\n"
+
+    def arrive_home(self) -> str:
+        self.location = self.home_location
+        return f"{self.name} rides home"
+
+
+class Car:
```
Oleksl888: Would make sense to put it in a separate module

app/customer.py
```diff
@@ -0,0 +1,55 @@
...
+    @classmethod
+    def change_fuel_price(cls, price: Decimal) -> None:
+        cls.FUEL_PRICE = Decimal(str(price))
+
+    def calculate_cost_distance(
+            self,
+            customer: Customer,
+            shop: object
+    ) -> Decimal:
+        sh_cord = (Decimal(shop.location[0]), Decimal(shop.location[1]))
```
Oleksl888: I would make a separate class for Location and define a method to calculate distance inside of it

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+from decimal import Decimal
+
+from app.customer import Customer, Car
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    with open("app/config.json", "r") as file:
+        json_data = json.load(file)
+    Car.change_fuel_price(json_data["FUEL_PRICE"])
```
Oleksl888, line 10: This is nice way to handle it. I guess you could also make it more simple if you make a separate module for getting json data and then pass this constant straight to class definition in Car, But you can keep it as is - it works

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    customers = [
```
Oleksl888: you can optimize this by having customers created and do all the shop trip related things in one loop. So you don't have to iterate over same data twice

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+                f"{customer.name}'s trip to the "
+                f"{shop.name} costs {all_cost}"
+            )
+            if all_cost < min_cost:
+                min_cost = all_cost
+                shop_index = index
+        if min_cost <= customer.money:
+            purchase_flag = True
+        if purchase_flag:
+            print(customer.arrive_at_shop(shops[shop_index]))
```
Oleksl888: i think all this prints and function calls would rather be put in Customer class as a method because all of this actions can be done within Customer instance - you can call it `do_shopping` or something.  Having so many prints inside your main() makes it look messy - it's better to organize this logic

app/shop.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class Shop:
+    def __init__(self, name: str, location: list, products: dict) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def calculate_purchase(self, customer: Customer) -> Decimal:
+        total_product_cost = Decimal("0")
```
Oleksl888: can you do sum on a generator expression or list comprehension here?

app/shop.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+            cost_for_products = decimal_quantity * decimal_shop_product_cost
+            if cost_for_products % 1 == 0:
+                cost_for_products = int(cost_for_products)
+            receipt += (
+                f"{quantity} {customer_product}s "
+                f"for {cost_for_products} dollars\n"
+            )
+            customer.product_cart[customer_product] = 0
+        receipt += total_cost_str + "See you again!\n"
```
Oleksl888, line 54: i like that you collected data and then print it all at once instead of making 10 prints :bow:

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #608, opened 2024-08-06
Solution Version 1.00

OleksiukStepan:


app/customer.py
```diff
@@ -0,0 +1,94 @@
...
+        return cheapest_shop
+
+    def get_shop_trip_cost(self, shop: Shop) -> None:
+        distance_price = self.get_distance_price(shop.location)
+        products_price = self.get_product_price(shop.products)
+        cost_of_trip = round(distance_price + products_price, 2)
+        print(f"{self.name}'s trip to the {shop.name} costs {cost_of_trip}")
+        self.shop_trips[cost_of_trip] = shop
+
+    def get_distance_price(self, shop_location: dict) -> float:
```

app/main.py
```diff
@@ -1,3 +1,28 @@
...
+
+    Car.fuel_price = config_data["FUEL_PRICE"]
+    customers_dict, shops_dict = config_data["customers"], config_data["shops"]
+    customers = [
+        Customer.get_customer_from_dict(customer)
+        for customer in customers_dict
+    ]
+    shops = [Shop.get_shop_from_dict(shop) for shop in shops_dict]
+
+    for customer in customers:
```
Oleksl888: Three for loops over the same data is a bit too much. You can reduce it to two loops for efficeincy. I would recommend to create Shops first, and then iterate over same data set to create customers, cars and do other things required

app/customer.py
```diff
@@ -0,0 +1,94 @@
...
+from app.car import Car
+from app.shop import Shop
+
+
+class Customer:
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: dict,
```
Oleksl888: I think you could implement a class Location that will have coords as parameters on init and will have calculate_distance method. This will structure your code better

app/shop.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+import datetime
+
+
+class Shop:
+    def __init__(
+            self,
+            name: str,
+            location: dict,
+            products: dict
```
Oleksl888: Product could also be presented as separate class and have method calcuclate_price

Oleksl888 (changes requested):

app/customer.py
OleksiukStepan: I also had doubts about this method. Now it’s clear :) Thanks!

LiudmylaKulinchenko (approved):
Looks good 🌸

PR #609, opened 2024-08-09
Function implemented

danilsiv:


app/car.py
```diff
@@ -0,0 +1,31 @@
+from dataclasses import dataclass
+from app.shop import Shop
+from app.customer import Customer
+
+
+@dataclass
```
EdAlekseiev: Why do you use dataclass here and for other classes?

EdAlekseiev (changes requested):

app/car.py
danilsiv: for simplify the syntax

vasylhnatiuk (approved):

PR #611, opened 2024-08-11
solution

BastovOleksandr:


app/customer.py
```diff
@@ -0,0 +1,66 @@
+from app.car import Car
+
+
+class Customer:
+    def __init__(
+        self,
+        name: str,
+        product_cart: dict,
+        location: list,
+        money: float,
```
oliasymak24: For money usually better to use Decimal module

app/customer.py
```diff
@@ -0,0 +1,66 @@
...
+        car: dict
+    ) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.money = money
+        self.car = Car(**car)
```
oliasymak24: cars will be better

app/customer.py
```diff
@@ -0,0 +1,66 @@
...
+            fuel_price: float
+    ) -> float:
+        distance = ((self.location[0] - to_location[0]) ** 2
+                    + (self.location[1] - to_location[1]) ** 2) ** 0.5
+
+        return distance * car.fuel_consumption * 0.01 * fuel_price * 2
+
+    def count_prod_cost(self, prices: dict) -> float:
+        return sum(
+            a * b for a, b in zip(self.product_cart.values(), prices.values())
```
oliasymak24: What is `a` and `b`, choose better variable names

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,3 +1,36 @@
...
+
+
+def shop_trip() -> None:
+    with open(f"app{os.sep}config.json", "r") as file:
+        trip_data = json.load(file)
+
+    customers, shops = [], []
+    fuel_price = Decimal(str(trip_data["FUEL_PRICE"]))
+
+    for cust_characteristics in trip_data["customers"]:
```
Oleksl888: Three for loops over the same data is a bit too much. You can reduce it to two loops for efficeincy. I would recommend to create Shops first, and then iterate over same data set to create customers, cars and do other things

app/main.py
```diff
@@ -1,3 +1,36 @@
...
+        cust_characteristics["car_characteristics"] = (
```
Oleksl888: cust_charactersitics is a very weird name for a variable. customer_data or similar would be better

app/customer.py
```diff
@@ -0,0 +1,74 @@
...
+        cheapest_shop = min(all_shops_total, key=all_shops_total.get)
+        return cheapest_shop, all_shops_total[cheapest_shop]
+
+    def count_fuel_cost(
+            self,
+            to_location: list,
+            car: Car,
+            fuel_price: Decimal
+    ) -> Decimal:
+        distance = (((self.location[0] - to_location[0]) ** 2
```
Oleksl888: I think you could implment a class Location that will have coords as parameters on init and will have calculate_distance method that will do this block of code. This will structure your code better imo

app/customer.py
```diff
@@ -0,0 +1,74 @@
...
+                     quantize(Decimal("0.01"), rounding=ROUND_HALF_UP))
+
+            all_shops_total[shop] = total
+
+            print(f"{self.name}'s trip to the {shop.name} costs {total}")
+
+        cheapest_shop = min(all_shops_total, key=all_shops_total.get)
+        return cheapest_shop, all_shops_total[cheapest_shop]
+
+    def count_fuel_cost(
```

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #613, opened 2024-08-13
Solution Politov

Alexandr-Politov:
Update readme

### DIFF
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from decimal import Decimal
 
 
 @dataclass
@@ -6,5 +7,8 @@ class Customer:
     name: str
     product_cart: dict
     location: list[int]
-    money: int
+    money: Decimal
     car: dict
+
+    def __post_init__(self) -> None:
+        self.money = Decimal(str(self.money))
--- a/app/customer.py
+++ b/app/customer.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
 from decimal import Decimal
 
+from app.shop import Shop
+
 
 @dataclass
 class Customer:
@@ -12,3 +14,11 @@ class Customer:
 
     def __post_init__(self) -> None:
         self.money = Decimal(str(self.money))
+
+    def calc_dist(self, shop: Shop) -> Decimal:
+        return Decimal((((shop.location[0] - self.location[0]) ** 2
+                         + (shop.location[1] - self.location[1]) ** 2) ** 0.5))
+
+    def calc_trip_cost(self, shop: Shop, fuel_price: Decimal) -> Decimal:
+        return (self.calc_dist(shop) * Decimal(self.car["fuel_consumption"])
+                / Decimal("100") * fuel_price)
--- a/app/main.py
+++ b/app/main.py
@@ -60,11 +60,11 @@ def shop_trip() -> None:
             print("Date: 04/01/2021 12:33:41")
             print(f"Thanks, {customer.name}, for your purchase!")
             print("You have bought:")
-            print(f"{customer.product_cart['milk']} "
+            print(f"{customer.product_cart["milk"]} "
                   f"milks for {min_milk_cost} dollars")
-            print(f"{customer.product_cart['bread']} "
+            print(f"{customer.product_cart["bread"]} "
                   f"breads for {int(min_bread_cost)} dollars")
-            print(f"{customer.product_cart['butter']} "
+            print(f"{customer.product_cart["butter"]} "
                   f"butters for {min_butter_cost} dollars")
             print(f"Total cost is {min_milk_cost + min_bread_cost
                                    + min_butter_cost} dollars")
--- a/app/main.py
+++ b/app/main.py
@@ -4,7 +4,7 @@ from app.customer import Customer
 from app.shop import Shop
 
 
-with open("config.json", "r") as json_file:
+with open("app/config.json", "r") as json_file:
     data_dict = json.load(json_file)
 
 FUEL_PRICE = data_dict["FUEL_PRICE"]
--- a/app/main.py
+++ b/app/main.py
@@ -60,11 +60,11 @@ def shop_trip() -> None:
             print("Date: 04/01/2021 12:33:41")
             print(f"Thanks, {customer.name}, for your purchase!")
             print("You have bought:")
-            print(f"{customer.product_cart["milk"]} "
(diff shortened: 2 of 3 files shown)


app/customer.py
oliasymak24, line 1: └── py-shop-trip
  └── app
    ├── car.py
    ├── customer.py
    ├── main.py
    └── shop.py
    
    This is a structure from the task, you to add car

app/customer.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Customer:
+    name: str
+    product_cart: dict
+    location: list[int]
+    money: int
```
oliasymak24: For money usually better ti yse Decimal module

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            print(f"{customer.name}'s trip to the "
+                  f"{shop.name} costs {total_cost}")
+
+            if total_cost <= min_total_cost:
+                min_total_cost = total_cost
+                min_milk_cost = milk_cost
+                min_bread_cost = bread_cost
+                min_butter_cost = butter_cost
+                min_shop = shop.name
+
```
oliasymak24: Make it dry

oliasymak24 (changes requested):

app/customer.py
Alexandr-Politov: this is just one of the opthions. It is not 'must have', just recomended.
must have only 2 classes and modules
image

app/main.py
Alexandr-Politov: Sorry. I don't know how to "dry" this part. I think that all of this variable definitions are needed and located where needed.

app/customer.py
Alexandr-Politov: Done. Changed in customer class and in main.py.

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+                                 + (shop.location[1]
+                                    - customer.location[1]) ** 2) ** 0.5))
+            fuel_cost = (distance * Decimal(customer.car["fuel_consumption"])
+                         / Decimal("100") * FUEL_PRICE) * 2
+            milk_cost = (Decimal(str(customer.product_cart["milk"]))
+                         * Decimal(str(shop.products["milk"])))
+            bread_cost = (Decimal(str(customer.product_cart["bread"]))
+                          * Decimal(str(shop.products["bread"])))
+            butter_cost = (Decimal(str(customer.product_cart["butter"]))
+                           * Decimal(str(shop.products["butter"])))
```
LiudmylaKulinchenko: What if you'll have more products with time? For example, you'll need 10 more products. Or even 100, sinse there're a lot of products that could be sold in a shop. You're gonna need to write lines of code to cover it
You should consider it right from the start. To make it dry you should fix architecture. Create a Product class to cover all products creation and calculate it's cost

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+            print(f"""{customer.name} rides to {min_shop}\n
+Date: 04/01/2021 12:33:41
+Thanks, {customer.name}, for your purchase!
+You have bought:
+{customer.product_cart["milk"]} milks for {min_milk_cost} dollars
+{customer.product_cart["bread"]} breads for {int(min_bread_cost)} dollars
+{customer.product_cart["butter"]} butters for {min_butter_cost} dollars
+Total cost is {min_milk_cost + min_bread_cost + min_butter_cost} dollars
+See you again!\n\n{customer.name} rides home
+{customer.name} now has {customer.money - min_total_cost} dollars\n""")
```
LiudmylaKulinchenko: Why do you need it? It will be calculated and written each time you import this file
It shouldn't behave like that. This logic must be in functions

app/main.py
LiudmylaKulinchenko: Move serving customer logic to a shop class. You have a lot going on in this function, make sure to follow SRP

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+                product, customer.product_cart[product]
+            )
+            print(
+                f"{customer.product_cart[product]} {product}s for {prod_cost} "
+                f"dollars"
+            )
+        total_cost = min_shop.calc_total_cost(customer.product_cart)
+        print(f"""Total cost is {total_cost} dollars
+See you again!\n\n{customer.name} rides home
+{customer.name} now has {customer.money - min_total_cost} dollars\n""")
```
LiudmylaKulinchenko, line 62: this will be written each time you import this file. You should put it in function as well

app/shop.py
```diff
@@ -0,0 +1,23 @@
...
+        if product_cost % 1 == 0:
+            return product_cost.quantize(Decimal("0"))
+        return product_cost
+
+    def calc_total_cost(self, product_cart: dict) -> Decimal:
+        total_cost = Decimal("0")
+        for product in product_cart:
+            total_cost += (Decimal(str(self.products[product]))
+                           * Decimal(str(product_cart[product])))
+        return total_cost
```
LiudmylaKulinchenko: Are you sure you need all these type conversions?

LiudmylaKulinchenko: It can be written in one line using sum

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            total_cost = fuel_cost + products_price
+            total_cost = total_cost.quantize(Decimal("0.01"),
+                                             rounding=ROUND_HALF_UP)
+
+            print(f"{customer.name}'s trip to "
+                  f"the {shop.name} costs {total_cost}")
+
+            if total_cost <= min_total_cost:
+                min_total_cost = total_cost
+                min_shop = shop
```
LiudmylaKulinchenko, line 39: I would suggest to create separate function to calculate total cost as well (directly in main or in Customer class), but overall structure looks better!

app/shop.py
LiudmylaKulinchenko: Are you sure these methods should be in Shop, not in Customer? Total cost will be different for each customer depending on their parameters

LiudmylaKulinchenko (changes requested):

app/shop.py
Alexandr-Politov: It would be different to each shop as well depending on customer's list of products.

Alexandr-Politov: done

oliasymak24 (approved):

PR #614, opened 2024-08-14
implement py-shop-trip

ValeriiaKyr:


app/customer.py
```diff
@@ -0,0 +1,116 @@
...
+class Customer:
+
+    list_customer = []
+
+    def __init__(
+            self,
+            name: str,
+            product_cart: dict,
+            location: list,
+            money: int,
```
oliasymak24: For money usually better to use Decimal module

app/shop.py
```diff
@@ -0,0 +1,33 @@
...
+    ) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def __str__(self) -> str:
+        return self.name
+
+    @staticmethod
+    def create_shop() -> None:
```
oliasymak24: It must be classmethod

app/customer.py
```diff
@@ -0,0 +1,116 @@
...
+            car: Car
+    ) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.money = money
+        self.car = car
+
+    @staticmethod
```
oliasymak24: Classmethod

oliasymak24 (commented):
Your tests were failed, check everything once again

oliasymak24 (changes requested):
)

app/car.py
```diff
@@ -0,0 +1,25 @@
...
+class Car:
+
+    cars_dict = {}
+
+    def __init__(self, brand: str, fuel_consumption: float) -> None:
+        self.brand = brand
+        self.fuel_consumption = fuel_consumption
+
+    @staticmethod
+    def create_car() -> None:
```
Oleksl888: you don't need to read json in every class that you create. Would be more optimal to read it in a spearate module or in main,py and then pass data from there to your classes

app/main.py
```diff
@@ -1,3 +1,35 @@
-def shop_trip():
-    # write your code here
-    pass
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
```
Oleksl888, line 10: I'd recommend to use data from json file here instead of using it in factory methods in each of your classes. Optimally you will need two iterations over data in json file. First iteration would create a list of shops, and second will create Customers, Cars, do all the calculations and print results. You don't really need static or class methods to create instances of your class - you can simply do like Customer(**data)

app/customer.py
```diff
@@ -0,0 +1,121 @@
...
+                product_cart=custom.get("product_cart"),
+                location=custom.get("location"),
+                money=custom.get("money"),
+                car=Car.cars_dict[custom.get("name")],
+            )
+            Customer.list_customer.append(customer)
+
+    def distance_to_shop(self, some_shop: Shop) -> float:
+        config_path = os.path.join(os.path.dirname(__file__), "config.json")
+        with open(config_path, "r") as file_cost_fuel:
```
Oleksl888: Again, no need to read json on every method

app/customer.py
```diff
@@ -0,0 +1,121 @@
...
+            data = json.load(file_cost_fuel)
+
+        fuel_cost = data.get("FUEL_PRICE")
+
+        distance_c = self.location
```
Oleksl888: I would recommendto implement a class Location that will have coords as parameters on init and will have calculate_distance method that will do this block of code. This will structure your code better imo

app/customer.py
```diff
@@ -0,0 +1,121 @@
...
+            customer = Customer(
+                name=custom.get("name"),
+                product_cart=custom.get("product_cart"),
+                location=custom.get("location"),
+                money=custom.get("money"),
+                car=Car.cars_dict[custom.get("name")],
+            )
+            Customer.list_customer.append(customer)
+
+    def distance_to_shop(self, some_shop: Shop) -> float:
```

app/customer.py
```diff
@@ -0,0 +1,121 @@
...
+        cost_trip = round(
+            ((distance * self.car.fuel_consumption) / 100) * fuel_cost,
+            2
+        )
+
+        return cost_trip
+
+    def will_cost(self, one_shop: Shop) -> float:
+
+        cost_milk = (self.product_cart.get("milk")
```
Oleksl888: you should calculate cost of products in a loop and do not hardcode names of products

app/customer.py
```diff
@@ -0,0 +1,121 @@
...
+        cost_breads = (self.product_cart.get("bread")
+                       * one_shop.products.get("bread"))
+        cost_butter = (self.product_cart.get("butter")
+                       * one_shop.products.get("butter"))
+        total_cost = cost_milk + cost_breads + cost_butter
+
+        return total_cost
+
+    def create_check(self, one_shop: Shop) -> None:
+        cost_milk = (self.product_cart.get("milk")
```
Oleksl888: same here too much repetition that could have been avoided using cycles and there should be no hardcoded names for products

Oleksl888 (changes requested):
To sum up:
you need to rethink entire concept of your code.
1. Refactor your main so you can read json only once and then use this data to create your classes
2. Try to finish this task in two for loops over initial data for better efficiency
3. Refactor classes - offload some methods from Customer class, add additional class Location if you want to improve structure of your code
4. Class and static methods for creating instances can be kept but they are really unnecessary here - you can simply create objects directly by passing args.
5. Avoid repetition in code - make it DRy and use cycles

oliasymak24 (approved):

PR #615, opened 2024-08-14
a bit like knights... test passed

Narberal90:


app/data.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class Data:
+    data = None
+
+    @staticmethod
+    def load_config(filename: json) -> dict:
+        with open(filename, "r") as file:
+            info = json.load(file)
+            return info
```
oliasymak24: Why you need a class for this?
You have only one method which is anyway static

app/customer.py
```diff
@@ -0,0 +1,47 @@
...
+        self.product_cart = data.get("product_cart")
+        self.location = data.get("location")
+        self.money = data.get("money")
+        self.car = Car(
+            data.get("car").get("brand"),
+            data.get("car").get("fuel_consumption"))
+
+    def calculate_cost(self, shop: Shop) -> float:
+        x, y = self.location
+        x1, y1 = shop.location
```
oliasymak24: not good names for variables)

app/customer.py
```diff
@@ -0,0 +1,47 @@
...
+class Customer:
+    group_customers: list["Customer"] = []
+
+    def __init__(
+        self, **data
+    ) -> None:
+        self.name = data.get("name")
+        self.product_cart = data.get("product_cart")
+        self.location = data.get("location")
+        self.money = data.get("money")
```
oliasymak24: Convet it to Decimal

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+    with open("app/config.json", "r") as file:
+        Data.data = json.load(file)
+
+    Customer.generate_unit()
+    Shop.generate_shop()
+
+    for customer in Customer.group_customers:
+        print(f"{customer.name} has {customer.money} dollars")
+
+        shop_choice = [None, float("inf")]
```
oliasymak24: Why you need a list here?
You just use it everywhere like two separate components

oliasymak24 (changes requested):

MNDonut (approved):

PR #640, opened 2024-10-09
solution

dmytrik:


app/classes/car.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+@dataclass
+class Car:
+    brand: str
+    fuel_consumption: float
+
+    def calculate_cost_way(self, way: float, fuel_price: float) -> float | int:
+        result = (way / 100) * self.fuel_consumption * fuel_price
+        return result
```
Nattalli: can it be done in one line?

Nattalli (approved):
good job!

app/classes/car.py
dmytrik: it was in one line, but i changed logic many times and forgot exchange

PR #642, opened 2024-10-16
Solution

vladyslav-tmf:


app/customer.py
```diff
@@ -0,0 +1,47 @@
...
+            (self.location[0] - shop_location[0]) ** 2
+            + (self.location[1] - shop_location[1]) ** 2
+        )
+
+    def calculate_total_trip_cost(self,
+                                  shop: Shop,
+                                  fuel_price: float) -> float:
+        distance = self.distance_to(shop.location)
+        fuel_cost_to_shop = self.car.calculate_fuel_cost(distance, fuel_price)
+        fuel_cost_to_home = self.car.calculate_fuel_cost(distance, fuel_price)
```
Mxbely: fuel cost to shop == fuel cost to home, you avoid using calculate_fuel_cost twice

Mxbely (approved):
I think frequently use "noqa" isn't good, you can use round() for round to some decimal

PR #670, opened 2024-12-24
'Solution'

dottfmar:


4ikkamony (approved):
Great!
Nice one with format_value method for those prices in Shop, should've done so myself

PR #675, opened 2024-12-25
Solution

VovaGld:


4ikkamony (approved):
Nice, I especially approve of the use of factory methods.
IMHO, main.py does too much work, most of the calculations could be moved to the classes. Consider it)

PR #678, opened 2024-12-27
Solution

Dogober:


app/car.py
```diff
@@ -0,0 +1,7 @@
+class Car:
+    def __init__(self, brand: str, fuel_consumption: float) -> None:
+        self.brand = brand
+        self.fuel_consumption = fuel_consumption
+
+    def fuel_needed_for_trip(self, distance: float | int) -> float | int:
```
L1nk3rrr: it will return float every time, because you have division

app/customer.py
```diff
@@ -0,0 +1,53 @@
...
+            self,
+            shop_location: list[int]
+    ) -> float | int:
+        x1, y1 = self.location
+        x2, y2 = shop_location
+        return math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2) * 2
+
+    def purchases_cost(self, products: dict[str, int]) -> float | int:
+        costs = 0
+        for product in self.product_cart.keys():
```
L1nk3rrr: Don't use `keys` simply do `for product in self.product_cart`

app/customer.py
```diff
@@ -0,0 +1,53 @@
...
+            costs += self.product_cart[product] * products[product]
+
+        return costs
+
+    def go_home(self) -> None:
+        print(f"{self.name} rides home")
+        print(f"{self.name} now has {self.money} dollars")
+        print("")
```
L1nk3rrr: you can call empty print without empty string

app/customer.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def go_buy(self, shop_name: str, costs_amount: float | int) -> None:
+        self.money -= costs_amount
+        print(f"{self.name} rides to {shop_name}")
+        print("")
```
L1nk3rrr: you can call empty print without empty string

app/shop.py
```diff
@@ -0,0 +1,39 @@
...
+            location: list[int],
+            products: dict[str, int]
+    ) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def print_check(self, customer: Customer) -> None:
+        current_time = datetime.datetime.now()
+        current_time = current_time.strftime("%d/%m/%Y %H:%M:%S")
```
L1nk3rrr: you can do it in one line

`datetime.datetime.now().strftime("%d/%m/%Y %H:%M:%S")`

L1nk3rrr (changes requested):
Well done! Review my comment to improve your code!

L1nk3rrr (approved):
Nice job!

PR #680, opened 2025-01-04
Solution

IvankaKuzin:


app/car.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, brand: str, fuel_consumption: Decimal) -> None:
+        self.brand = brand
+        self.fuel_consumption = Decimal(fuel_consumption)
+
+    def count_fuel_consumption(
+        self, fuel_price: Decimal, length_way: Decimal
+    ) -> Decimal:
+        return (
+            ((Decimal(length_way) / 100) * self.fuel_consumption)
+            * Decimal(fuel_price)
```
L1nk3rrr: length_way, fuel_price - must be a Decimal from your function, why you transform it again?

app/car.py
```diff
@@ -0,0 +1,16 @@
+from decimal import Decimal
+
+
+class Car:
+    def __init__(self, brand: str, fuel_consumption: Decimal) -> None:
+        self.brand = brand
+        self.fuel_consumption = Decimal(fuel_consumption)
```
L1nk3rrr: Same, you are waiting for Decimal and make it Decimal here?

L1nk3rrr (commented):
Well done! Minor point for improvement - try to convert your data only when you need this, cause I see that you have places where you are waiting for Decimal, but convert it to Decimal in method/function AND also basically it's already converted.

L1nk3rrr (changes requested):
Well done! Minor point for improvement - try to convert your data only when you need this, cause I see that you have places where you are waiting for Decimal, but convert it to Decimal in method/function AND also basically it's already converted.

L1nk3rrr (approved):
Well done!

PR #693, opened 2025-02-04
solution

Bleit1:
изображение

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,36 @@
-def shop_trip():
-    # write your code here
-    pass
+from os import path
+
+from app.data_loader import DataLoader
+from app.customer import Customer
+from app.shop import Shop
+
+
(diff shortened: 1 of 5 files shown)


app/main.py
```diff
@@ -1,3 +1,36 @@
...
-    pass
+from os import path
+
+from app.data_loader import DataLoader
+from app.customer import Customer
+from app.shop import Shop
+
+
+def shop_trip() -> None:
+    data = DataLoader(path.join("app", "config.json"))
```
L1nk3rrr: Why do you need DataLoader if you create all your data here, in short_trip

L1nk3rrr (changes requested):
Tests are not passed, check it before sending to review

Nattalli (changes requested):
fix tests

PR #709, opened 2025-03-07
Completed task shop-trip

Yurkiv313:


app/main.py
```diff
@@ -1,3 +1,93 @@
...
+            print(f"{customer.name} rides home")
+            print(
+                f"{customer.name} now has "
+                f"{customer.money - all_sum_trip[0][0]} dollars\n"
+            )
+        else:
+            print(
+                f"{customer.name} doesn't have enough money"
+                f" to make a purchase in any shop"
+            )
```
Boyok1337: Too long func, spliit to diffferent methods, try to add loading data file as example, then the corresponding prints should be like methods in the corresponding class. 
Simplify logic and run it via methods

app/car.py
```diff
@@ -0,0 +1,11 @@
+import dataclasses
+
+
+@dataclasses.dataclass
```
Boyok1337: If you use only `dataclass` from `dataclasses`, so import only this

Boyok1337 (changes requested):

app/main.py
```diff
@@ -1,3 +1,137 @@
...
+    )
+    return chosen_shop
+
+
+def print_way_to_shop(customer: Customer, chosen_shop: Shop) -> None:
+    print(f"{customer.name} rides to {chosen_shop.name}\n")
+
+
+def print_way_to_home(customer: Customer) -> None:
+    print(f"{customer.name} rides home")
```
Boyok1337: It's better, but you should have implemented the prints as class methods

app/main.py
```diff
@@ -1,3 +1,137 @@
...
+
+
+def choose_cheapest_shop(
+    shops: List[Shop], all_sum_trip: List[Tuple[float, str]]
+) -> Shop:
+    chosen_shop_name = all_sum_trip[0][1]
+    chosen_shop = next(
+        shop for shop in shops if shop.name == chosen_shop_name
+    )
+    return chosen_shop
```
Boyok1337: These should also be methods of the corresponding class, not just functions

app/main.py
```diff
@@ -1,3 +1,137 @@
...
+            customer.money -= all_sum_trip[0][0]
+            print(f"{customer.name} now has {customer.money} dollars\n")
+        else:
+            print(
+                f"{customer.name} doesn't have enough money"
+                f" to make a purchase in any shop"
+            )
+
+
+shop_trip()
```
Boyok1337: To simply call a function in this file, you should use the corresponding `main check `construct

Boyok1337 (changes requested):

Y-Havryliv (approved):

PR #722, opened 2025-03-22
Solution

HappyDen08:


app/car.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Car:
+    brand: str
+    f_consumption: float
```
Boyok1337, line 7: Use full and clear names

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+            Car(customer["car"]["brand"],
+                customer["car"]["fuel_consumption"])
+        ))
+
+    for shop in db_data["shops"]:
+        shops[shop["name"]] = Shop(
+            shop["name"],
+            shop["location"],
+            shop["products"]
+        )
```
Boyok1337, line 32: Try to separate it to other func like load_data

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+            print(line)
+
+        print(f"Total cost is {total} dollars")
+        print("See you again!")
+        print()
+        print(f"{customer.name} rides home")
+        print(f"{customer.name} now has {customer.money} dollars")
+        customer.location = customer.home_location
+
+        print()
```
Boyok1337, line 77: Divide the logical part into functions and start this functions in main func

Boyok1337 (changes requested):

PR #758, opened 2025-07-16
Solution

viannik:


Arch0998 (approved):

viannik:
@mateacademy-ai-mentor I use the class objects as keys and my own hash because it makes sense for the task. A store is unique when it has a unique name and location. Is it worth changing anything in this case?

PR #765, opened 2025-07-19
Develop

Plovuezz:


app/customer.py
```diff
@@ -0,0 +1,120 @@
...
+        dy = shop.location[1] - self.location[1]
+        distance_km = sqrt(dx ** 2 + dy ** 2)
+
+        fuel_cost = self.car.fuel_consumption * distance_km / 100 * fuel_price
+
+        product_cost = sum([
+            shop.products[product] * self.product_cart[product]
+            for product in self.product_cart
+            if product in shop.products
+        ])
```
dimashevchukk: unnesecary list comprehension

app/customer.py
```diff
@@ -0,0 +1,120 @@
...
+        product_cost = {
+            product: [self.product_cart[product], shop.products[product]]
+            for product in self.product_cart
+            if product in shop.products
+        }
+        total = sum([
+            shop.products[product] * self.product_cart[product]
+            for product in self.product_cart
+            if product in shop.products
+        ])
```
dimashevchukk: unnesecary list comprehension

app/customer.py
```diff
@@ -0,0 +1,120 @@
...
+
+        print(f"Date: {current_date}")
+        print(f"Thanks, {self.name}, for your purchase!")
+        print("You have bought:")
+        for product in product_cost:
+            purchase = round(product_cost[product][0]
+                             * product_cost[product][1], 1)
+            if purchase == 3.0:
```
dimashevchukk, line 71: format any purchase

app/customer.py
```diff
@@ -0,0 +1,120 @@
...
+def get_data_from_json(filename: str) -> dict:
+    with open(filename, "r") as json_file:
+        data = json.load(json_file)
+
+    return data
+
+
+def process_data(data_dict: dict) -> tuple:
+
+    fuel_price = data_dict.get("FUEL_PRICE", None)
```
dimashevchukk, line 93: None is default param

dimashevchukk (changes requested):

app/customer.py
Plovuezz: @dimashevchukk cant understand what is the problem here, can you specify?

app/customer.py
Plovuezz: @dimashevchukk here i have an issue with this. For some reason when the result is 3 it is output as 3.0 and no formatting helps to solve this. So this was the only option

dimashevchukk: @Plovuezz understandable, i had the same one)
look mine or somebody else's solution, in my code formatting looks like
formated_price = f"{product_price:.2f}".rstrip("0").rstrip(".")

dimashevchukk (approved):

PR #803, opened 2025-11-04
'Solution'

Vorongor:


app/customer.py
```diff
@@ -0,0 +1,85 @@
...
+    ) -> None:
+        self.name = name
+        self.product_cart = product_cart
+        self.location = location
+        self.money = money
+        self.car = Car.setup_car(car, location)
+        self.cheapest_shop = None
+        self.cheapest_cost = 0
+
+    def set_location(self,
```
DonDmytron: What do you think about using property and setter in this simple case ?

app/customer.py
```diff
@@ -0,0 +1,85 @@
...
+        ) / 100
+
+    def preparation(self,
+                    shops: list[Shop],
+                    fuel_price: float) -> None:
+        """Compare shops, print costs and choose cheapest."""
+        print(f"{self.name} has {self.money} dollars")
+
+        results = []
+        for shop in shops:
```
DonDmytron: I think this logic would be better in the Shop class as a separate method, and list comprehension could be used here.

app/car.py
```diff
@@ -0,0 +1,34 @@
+import math
+
+
+class Car:
+    """Car that calculates travel cost based on fuel consumption."""
+
+    def __init__(self,
```
DonDmytron: Please write declarations of all methods following the example of the init method of the Customer class.

DonDmytron (changes requested):

DonDmytron (approved):
GJ

PR #804, opened 2025-11-04
Solution

EduardoMoen:


meruyme (approved):
Congratulations, Eduardo! Your solution is correct, so I'm approving it. Just a tip: I noticed that you created a `move_to` method in the Customer class, but didn't use it anywhere. I think this method would be useful in the `shop_trip` function, when the customer is changing location between the shop and his home.

PR #889, opened 2026-04-18
solution

Reverad:


app/customer.py
```diff
@@ -0,0 +1,56 @@
+import math
+
+from app.shop import Shop
+
+
+class Customer:
+    def __init__(self, dct: dict, fuel_price: float) -> None:
+        self.name = dct["name"]
```
AnyoneClown, line 8: Use better naming for variables

AnyoneClown (approved):
LGTM

PR #910, opened 2026-06-03
Solution

AndriiOne:


app/shop.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, name: str, location: list, products: dict) -> None:
+        self.name = name
+        self.location = location
+        self.products = products
+
+    def get_product_cost(self, product_cart: dict) -> float:
+        total_cost = 0
+        for product, count in product_cart.items():
+            total_cost += count * self.products[product]
+        return total_cost
```
vsmutok: Perhaps you could use a list comprehension here)

app/shop.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def print_recipe(self, customer_name: str, product_cart: dict) -> None:
+        print(f"Date: {datetime.datetime.now().strftime('%d/%m/%Y %H:%M:%S')}")
+        print(f"Thanks, {customer_name}, for your purchase!")
+        print("You have bought:")
```
vsmutok: If several `print()` statements are used consecutively, you can combine them into a single `print()` statement.

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    fuel_price = data["FUEL_PRICE"]
+
+    shops = []
+    for shop_data in data["shops"]:
+        shop_obj = shop.Shop(
+            name=shop_data["name"],
+            location=shop_data["location"],
+            products=shop_data["products"]
+        )
+        shops.append(shop_obj)
```
vsmutok:
```suggestion
    shops = [shop.Shop(**shop_data) for shop_data in data["shops"]]
```
List comprehension and unpacking seem more appropriate here 🙂

vsmutok (changes requested):

app/main.py
AndriiOne: Done. Refactored both loops into list comprehensions and used dictionary unpacking (**)

app/shop.py
AndriiOne: Done! Combined consecutive prints into a single print() statement using multi-line string implicit concatenation with \n

app/shop.py
AndriiOne: Done! Fixed it by using a generator expression inside sum() instead of a list comprehension to optimize memory usage

vsmutok (approved):
Great!

PR #930, opened 2026-08-06
Solution

swirtam:


swirtam:
@mateacademy-ai-mentor I was trying to avoid defining fuel price inside the class, so I can import json module and open configuration file only in main.py. I then set a value for class variable FUEL_PRICE in shop_trip function before creating any class attributes. As for shop.py and main.py, I can see both of them in section Files changed (5 files in total)

